PR #1, opened 2021-02-08
Initial rules about the catalog

nicolaferraro:
Some general rules for contributions to the official Kamelet catalog.

cc: @davsclaus , @oscerd , @astefanutti , @squakez , @doru1004 , @lburgazzoli


oscerd (approved):

astefanutti (approved):

README.md
```diff
@@ -1,6 +1,126 @@
...
+
+
+### Flow Code
+
+The Camel route that defines the behavior of the Kamelet **MUST** be provided in YAML flow syntax.
+The Kamelet **MAY** declare additional supporting routes that **MUST** be written in YAML syntax (to provide better support in all Camel subprojects).
+
+The code of a "source" Kamelet must send data to the `kamelet:sink` special endpoint. The code of a "sink" Kamelet must consume data from the special endpoint `kamelet:source`.
+
+The Kamelet **CAN** declare dependencies on Camel components using the syntax `camel:<component-name>` (e.g. `camel:telegram`).
```
davsclaus, line 88: For these CAN it would be good to have an example showing that, eg can we add another out of the box kamelet that does that.

Also the timer kamelet does not declare that it uses camel:timer

README.md
```diff
@@ -1,6 +1,126 @@
...
+For the time being, we'll accept only two kind of Kamelets:
+
+- **Sources**: Kamelets producing data that can be forwarded to any chosen destination. In the Camel jargon, a source can be used consumer-side.
+Kamelets belonging to this category **MUST** be marked with label: `camel.apache.org/kamelet.type=source`.
+- **Sinks**: Kamelets that accept data with a specific datashape and forward it to an external system. In the Camel jargon, a sink can be used producer-side.
+Kamelets belonging to this category **MUST** be marked with label: `camel.apache.org/kamelet.type=sink`.
+
+All Kamelets **MUST** provide a value for label `camel.apache.org/kamelet.type`.
+
+All Kamelets **MUST** declare an icon in the `camel.apache.org/kamelet.icon` annotation using the embedded URL `data:image` format. An icon annotation **CANNOT** contain a link to an external location. A Kamelet **SHOULD** use the specific `data:image/svg+xml;base64` format whenever it's possible.
```
davsclaus, line 33: Yes SVG icons are the best. I wonder if we have any way of previewing those icons from a PR. Or some hint. Maybe you can just copy that base64 into some online service that can render it.

And do we have a generic base64 icon for camel itself? But I would of course like to see specific icons per kamelet so they use twitter, telegram, salesforce logos/icons etc. But finding svg icons can sometime be tricky. You may have a bitmap icon, and if you convert that to SVG then it gets very bloated and big. So to find a natural clean svg icon is not always easy

README.md
```diff
@@ -1,6 +1,126 @@
...
+spec:
+  definition:
+    title: Timer Source
+    description: Produces periodic events with a custom payload
+    required:
+      - message
+    properties:
+      period:
+        title: Period
+        description: The interval between two events
```
davsclaus, line 57: (in milli seconds)

davsclaus (approved):

README.md
nicolaferraro: Yeah, I see your concerns... I thought it would make sense to have a web module generated for the kamelets in this repo and previewed by Netlify on each PR. This way we can either check the icon (but also the schema) in the browser and also include the Kamelet catalog in the main Camel website. Could it work?

squakez (approved):

PR #3, opened 2021-02-12
chore: add CI configuration

nicolaferraro:
This should setup a CI based on Github actions. I've added the test for the timer-source Kamelet.

We should find a good solution for easily manage credentials before we can test the more complex ones.


oscerd (approved):

nicolaferraro:
Let's merge this to enable actions

PR #6, opened 2021-02-16
chore: fix generated YAML and required props

nicolaferraro:


oscerd (approved):

davsclaus (approved):

davsclaus:
btw is the source code generated list of options in the same order as in the yaml file (eg so users may put the most important in the top).

nicolaferraro:
> btw is the source code generated list of options in the same order as in the yaml file (eg so users may put the most important in the top).

Well, currently it should be alphabetical with required fields first... But we can think about it..

nicolaferraro:
> > btw is the source code generated list of options in the same order as in the yaml file (eg so users may put the most important in the top).
> 
> Well, currently it should be alphabetical with required fields first... But we can think about it..

The problem is that it's a map that loses the order when unmarhsalled..

davsclaus:
> > > btw is the source code generated list of options in the same order as in the yaml file (eg so users may put the most important in the top).
> > 
> > 
> > Well, currently it should be alphabetical with required fields first... But we can think about it..
> 
> The problem is that it's a map that loses the order when unmarhsalled.. that's more a problem that happens e.g. on Kube, once you apply it you lose the order

ah okay A..Z is also fine, just that its the same. 

We had that problem earlier in core camel where a regen would shuffle the options. So its best to have a fixed sorting in the build process

PR #42, opened 2021-03-08
Added SFTP source Kamelet

### LINKED ISSUE #23: Add SFTP Source Kamelet

oscerd:
Fixes #23


sftp-source.kamelet.yaml
```diff
@@ -0,0 +1,68 @@
...
+    camel.apache.org/provider: "Apache Software Foundation"
+  labels:
+    camel.apache.org/kamelet.type: "source"
```
nicolaferraro, line 9: Should we add label camel.apache.org/kamelet.group=FTP in this and the FTP case?
 In case they belong to the same group we should also use the same icon, so that we can group them together in the visualization. Wdyt?

oscerd: Sounds good, do we want to use the ftp icon for both?

nicolaferraro: Yes, it will look odd at first, but we can change the website style later to add support for groups and fix it.

oscerd:
@nicolaferraro updated

nicolaferraro (approved):

PR #46, opened 2021-03-09
RabbitMQ source kamelet

### LINKED ISSUE #44: Add RabbitMQ source Kamelet

oscerd:
Fixes #44


rabbitmq-source.kamelet.yaml
```diff
@@ -0,0 +1,54 @@
...
+  flow:
+    from:
+      uri: "rabbitmq://{{exchangeName}}"
+      parameters:
+        password: "{{password}}"
+        username: "{{username}}"
+        addresses: "{{addresses}}"
+      steps:
+      - convert-body-to:
+          type: "java.lang.String"
```
nicolaferraro, line 53: Does it need this conversion? I guess it already produces binary data that can be sent over the wire, right?

oscerd: Yes, bytes only

nicolaferraro: Yeah, so I would send it directly to the sink

oscerd: Fixed

PR #51, opened 2021-03-10
Gmail Stream source kamelet

### LINKED ISSUE #19: Add Google Mail Stream Source Kamelet

oscerd:
Fixes #19


google-mail-stream-source.kamelet.yaml
```diff
@@ -0,0 +1,93 @@
...
+        title: Application name
+        description: Google Mail application name
+        type: string
+      markAsRead:
+        title: Mark as read
+        description: Mark the message as read once it has been consumed
+        type: boolean
+        x-descriptors:
+        - 'urn:alm:descriptor:com.tectonic.ui:checkbox'
+        default: false
```
nicolaferraro, line 61: Better setting the default to `true`, so we have a streaming behavior without duplicates by default. Wdyt?

oscerd: I'm ok, that is the camel catalog default, that's the reason

oscerd: done

PR #107, opened 2021-03-30
AWS S3 Sink Kamelet

claudio4j:
#112


oscerd (changes requested):
I think you should follow the approach we use in dropbox sink: https://github.com/apache/camel-kamelets/blob/master/dropbox-sink.kamelet.yaml#L51

And state in the description that the sink expects the header to be set.

claudio4j:
In camel-dropbox if there is no header to designate the filename, it uses the exchange id, but in camel-aws2-s3 there is no such default for filename. So, in this case I think requiring the filename is appropriate isn't ? Or should we set the exchange id if there is no CamelAwsS3Key header ?

oscerd:
> In camel-dropbox if there is no header to designate the filename, it uses the exchange id, but in camel-aws2-s3 there is no such default for filename. So, in this case I think requiring the filename is appropriate isn't ? Or should we set the exchange id if there is no CamelAwsS3Key header ?

I meant to say the same style as dropbox, by adding an otherwise step to set the header as exchange id. Like in the minio sink connector for example: https://github.com/apache/camel-kamelets/blob/master/minio-sink.kamelet.yaml#L60

aws-s3-sink.kamelet.yaml
```diff
@@ -0,0 +1,95 @@
...
+            steps:
+            - set-header:
+                name: CamelAwsS3Key
+                simple: "${header[ce-file]}"
+          otherwise:
+            steps:
+            - set-header:
+                name: CamelAwsS3Key
+                simple: "${exchangeId}"
+      - log: "s3 body - ${body}"
```
oscerd, line 86: I don't think this is needed, eventually a user can add a step with a log sink in the middle of a binding.

oscerd (changes requested):
Just a little thing.

aws-s3-sink.kamelet.yaml
claudio4j: Ahh, just forgot to remove it before pushing.

nicolaferraro (approved):

Issue #113, opened 2021-04-02
Add generic set-header action


claudio4j:
A kamelet is required to set the header in a kamelet binding, as many kamelets may set headers.

claudio4j:
As requested in the PR, lets add this type of kamelet to another repo.

PR #114, opened 2021-04-02
Kamelet set-header to reuse across kamelet bindings

claudio4j:
#113


oscerd (commented):
We need to find a better place for this kind of kamelets. Let's place this on hold until we have an agreement.

PR #144, opened 2021-04-14
PDF Action Kamelet

oscerd:
This is an example of action between a source and a sink. The PDF Action needs to have a sink after it to write the file somewhere (File, FTP, SFTP etc.)


nicolaferraro (approved):
Cool, we should start accepting actions like this one, now that 1.4.0 supports intermediate steps in bindings. The generated doc is not aligned, I'm going to create some issues for that

PR #154, opened 2021-04-20
AWS SQS Batch sink

### LINKED ISSUE #137: Add an AWS SQS Batch Sink Kamelet
With 3.9.0 we support to specify a separator for batch messages, so we can create a different kamelet for this particular scenario

oscerd:
Fixes #137


aws-sqs-batch-sink.kamelet.yaml
```diff
@@ -0,0 +1,66 @@
...
+spec:
+  definition:
+    title: AWS SQS Batch Sink
+    description: |-
+      Send Batch message to an AWS SQS Queue
+    required:
+      - queueNameOrArn
+      - accessKey
+      - secretKey
+      - region
```
nicolaferraro, line 19: batchSeparator also seems required

nicolaferraro (changes requested):

aws-sqs-batch-sink.kamelet.yaml
oscerd: done

PR #170, opened 2021-04-26
DNS IP Action

oscerd:
Related to #168


nicolaferraro (commented):
Since this modifies the body, can we also fallback to using the body as source ip when the domain header is not present? This way it can be used as simple body transformation from DNS to ip, like a "resolve action". Wdyt?

nicolaferraro (approved):

PR #174, opened 2021-04-26
Dns lookup Action Kamelet

### LINKED ISSUE #168: Add DNS ip, dig and lookup Kamelet actions

oscerd:
Fixes #168


dns-lookup-action.kamelet.yaml
```diff
@@ -0,0 +1,50 @@
...
+            steps:
+            - set-header:
+                name: dns.domain
+                simple: "${header[ce-domain-name]}"
+          otherwise:
+            steps:
+            - set-header:
+                name: dns.domain
+                simple: "${body}"
+      - marshal:
```
nicolaferraro, line 47: Seems that marshal and to are in reversed order, isn't it?

nicolaferraro (changes requested):

dns-lookup-action.kamelet.yaml
oscerd: Fixed. Wrong commit pushed.

nicolaferraro (approved):

PR #180, opened 2021-04-28
Infinispan Source Kamelet

oscerd:


infinispan-source.kamelet.yaml
```diff
@@ -0,0 +1,80 @@
...
+      - hosts
+      - username
+      - password
+      - hostAddresses
+    properties:
+      hosts:
+        title: Hosts
+        description: Specifies the host of the cache on Infinispan instance
+        type: String
+        default: 1000
```
nicolaferraro, line 34: Default seems not right. Also you may want to specify if multiple hosts should be e.g. comma separaterd or so

infinispan-source.kamelet.yaml
```diff
@@ -0,0 +1,80 @@
...
+        hosts: "{{hosts}}"
+        secure: "{{secure}}"
+        username: "{{username}}"
+        password: "{{password}}"
+        saslMechanism: "{{saslMechanism}}"
+        securityRealm: "{{securityRealm}}"
+        securityServerName: "{{securityServerName}}"
+        eventTypes: "{{eventTypes}}"
+      steps:
+        - to: kamelet:sink
```
nicolaferraro, line 80: Don't know, is the result already serialized?

nicolaferraro (changes requested):

infinispan-source.kamelet.yaml
oscerd: The result is the exchange with some headers exposing the event. I can make it a json if needed.

infinispan-source.kamelet.yaml
oscerd: copy paste error. I'll fix it.

oscerd:
I'm merging this one and I'll work on headers mapping for knative.

Issue #225, opened 2021-05-04
Avoid Kamelets Catalog to throw IOException


apupier:
It woudl be more convenient when using it.

I think it is reasonable that the list of kamelets is empty in the catalog in case of IOException

oscerd:
Sounds good. If you have time and will, it would be good to have a pr

apupier:
I can have a look at it

PR #226, opened 2021-05-04
Avoid IOException on constructor #225

apupier:
- when a kamelet file cannot be loaded, a log is provided but other
files still can load
- provided a test to ensure that we are not embedding some kamelet files
that are invalid and that we won't detect anymore with existing tests

Signed-off-by: Aurélien Pupier <apupier@redhat.com>


oscerd (approved):

oscerd:
Fixes #226

oscerd:
Thanks.

PR #265, opened 2021-05-10
Json serialize action

### LINKED ISSUE #244: Provide a json-deserialize-action and a json-serialize-action
Internally the deserialize action should deserialize to JSONNode, so that we can append transformations.

One simple use case where this is needed:

Start from a aws-sqs-source. It does not have a type known in advance, but the user knows there's json inside the sqs queue, so, adding a deserialize step from JSON will make the data available for subsequent actions.

The user can also append actions to serialize to other data formats.

oscerd:
Fixes #244


nicolaferraro (approved):

nicolaferraro:
There are some conflicts with the nav

PR #362, opened 2021-06-14
AMQP JMS source kamelet

orpiske:


oscerd (changes requested):
Seems to be ok, but we need to use the same naming convention.

Also, run mvn clean install, so it will be added to catalog.

And after that have a look  at script/generator to generate docs and script/validate to validate.

orpiske:
@oscerd one problem that I foresee with this one is that it does not have a free icon available. If we enforce the MUST have `camel.apache.org/kamelet.icon` annotation, this one might be delayed until we clear this out or come up with an icon for it.

oscerd:
You can use any icon related to messaging, it shouldn't be a logo or something related to amqp. So even a message icon is fine.

orpiske:
> You can use any icon related to messaging, it shouldn't be a logo or something related to amqp. So even a message icon is fine.

That helps. Let's use the QPid one.

oscerd (approved):
cc @nicolaferraro

nicolaferraro (approved):

PR #366, opened 2021-06-14
MS SQL Server Sink Kamelet

### LINKED ISSUE #364: Create MS SQL Server Sink Kamelet

oscerd:
Fixes #364


sqlserver-sink.kamelet.yaml
```diff
@@ -0,0 +1,89 @@
...
+    camel.apache.org/kamelet.group: "SQL"
+  labels:
+    camel.apache.org/kamelet.type: "sink"
+spec:
+  definition:
+    title: "Microsoft SQL Server Sink"
+    description: |-
+      Send data to a Microsoft SQL Server Database.
+
+      In you KameletBinding file you'll need to explicitly declare the MariaDB driver dependency, like in the following YAML snippet
```
nicolaferraro, line 19: Typo about maria DB. Think we can reference the driver directly here

nicolaferraro: There are also other refs to mariaDB

nicolaferraro (changes requested):
I think we should put here and in the other SQL connectors some instructions. e.g. specifying that the input is expected to be JSON (in spec->types), an example in the query field and telling how the json fields are mapped to parameter values

oscerd:
I'll rework the other connector by adding expecting type, query example and json mapping.

oscerd:
Let me merge and open other PRs.

PR #401, opened 2021-06-25
Enhance mongodb sink kamelet

claudio4j:
Adding more parameters and expected headers to control document
filtering, sorting, limits, skip, etc

### DIFF
--- a/mongodb-sink.kamelet.yaml
+++ b/mongodb-sink.kamelet.yaml
@@ -17,6 +17,10 @@ spec:
       Send documents to MongoDB.
 
       This Kamelet expects a JSON as body.
+
+      Properties you can as headers:
+
+      `db-upsert` / `ce-db-upsert` If the database should create the element if it does not exist. Boolean value.
     required:
       - hosts
       - collection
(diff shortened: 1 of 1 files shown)


oscerd:
To me a kamelet should be a pre-baked action. So it should be tailored just for one operation. If we need to add more operations they should be different kamelets. So i would focus on options related to insert operation (headers and endpoint params) and leave the operation as insert and static. If you think about the kafka connectivity for example, if you use the sink kamelet you'll go for inserting Kafka records in mongodb

claudio4j:
It's ok for me to add another mongodb-op-sink kamelet so users can specify which operation to run, for example a findAll operation with filter/sort/limit parameters, sounds appropriate for users who want to retrieve data from mongodb.

oscerd:
If they need to retrieve something from mongodb they should use a source kamelet

oscerd:
I think you should create specialized sink kamelet for more operations on the sink side, but I wouldn't add all the headers in one place. For the general sink, I would just add the writeConcern and createCollection properties and leave it as is. The aim of kamelets is to hide the Camel internals to the users.

claudio4j:
@oscerd In addition to those you mentioned I think db-upsert is useful too.

oscerd (approved):

PR #423, opened 2021-07-13
ci: tweak test workflow

lburgazzoli:


.github/workflows/test.yaml
```diff
@@ -87,8 +112,16 @@ jobs:
...
+    - name: Run Tests
+      run: |
+        # Configure install options
+        export KAMEL_INSTALL_BUILD_PUBLISH_STRATEGY=Spectrum
+        export KAMEL_INSTALL_REGISTRY=$KIND_REGISTRY
+        export KAMEL_INSTALL_REGISTRY_INSECURE=true
+
         echo "Running tests"
-        yaks test ./test
+        yaks run --logs --wait ./test
```
nicolaferraro, line 123: Does not seem to wait...
```suggestion
        yaks test ./test
```

lburgazzoli: yeah, it was an attempt to see if for some reason the cli was not working as expected
btw, yaks test does not seems to exists as a sub-command is that correct ?

lburgazzoli:
@nicolaferraro don't know if it is worth merging it, it was mainly an attempt to understand why it was not working.

oscerd:
Do we want to merge this one?

lburgazzoli:
don't think it is needed anymore

PR #429, opened 2021-07-15
Add slack sink kamelet

doru1004:
Add slack sink kamelet


oscerd (changes requested):
It makes sense to add more options as optional like for example userName and iconEmoji.

oscerd (changes requested):
LazystartProducer doesn't really make sense in this context. It's something internal. I would remove it. Thanks. Overall looks good

oscerd (approved):

nicolaferraro (approved):
Cool, thanks!

PR #444, opened 2021-07-26
Added Velocity-based templates for the Kamelets

orpiske:
Asking for some initial feedback for a new feature for Camel JBang. 

I'm working on the ability to generate Kamelets from templates. The idea is to make it easier for user to bootstrap their own kamelets. To do so, it uses the templates on this repository.


oscerd (approved):
Looks good as beginning.

oscerd:
I think we are good to go here @orpiske

orpiske:
> I think we are good to go here @orpiske

Thanks, I'm going to merge this one.

PR #456, opened 2021-08-03
Externalize the Kamelet binding template in the Kafka Source

orpiske:
Move most of the code that generates documentation for bindings into external templates. 

I plan to re-use them to support a new feature for Camel JBang that creates the file for the user.


orpiske:
I refreshed the documentation as part of this only to showcase that the PR did not result in significant changes to the generated documentation.

oscerd (approved):

oscerd:
Is the generation script and validation script still ok?

orpiske:
> Is the generation script and validation script still ok?

Yes. Both work just like before.

oscerd:
+1 to merge.

PR #512, opened 2021-09-16
Exclude explicit quotes for chuck-norris source

### LINKED ISSUE #511: Adding filter for exluding explicit content to chuck-norris-source
I love the chuck-norris-source, and would like to use them in demos, but a bit afraid of showing up a quote like in https://api.chucknorris.io/jokes/7xoo4cjutwu3sgmuelp2pw during a customer demo ;-P While not necessarily not funny, it might confuse the audience a bit.

Would it be possible to add a filter from a bad-word list to https://github.com/apache/camel-kamelets/blob/fb004b99f4a54305ea5ec55b52944f2a8512bd0e/chuck-norris-source.kamelet.yaml#L39-L42 and just retry if a word of the list is matched? My Camel foo is unfortunately not good enough to add the corresponding steps ...

rhuss:
A future extension could add a parameter to re-enable explicite qoutes,
but for now I think it is safer to disable to avoid offending
people when using the source in demos.

Fixes #511.

### DIFF
--- a/chuck-norris-source.kamelet.yaml
+++ b/chuck-norris-source.kamelet.yaml
@@ -14,7 +14,7 @@ spec:
   definition:
     title: "Chuck Norris Source"
     description: |-
-      Gets peridically Chuck Norris jokes from http://www.icndb.com/
+      Gets periodically Chuck Norris jokes from http://www.icndb.com/
     type: object
     properties:
       period:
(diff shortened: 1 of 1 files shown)


oscerd (approved):

oscerd:
Thanks

PR #583, opened 2021-11-25
Moved all the kamelet.yaml files to kamelets folder

### LINKED ISSUE #448: Move kamelets from root folder into a sub folder
There are soo many kamelets now it would be good to move them into their own sub folder, as we also have code for other stuff like the catalog api, and whatnot.

That helps the structure a bit, and also to look at the project on github as the readme file is shown sooner, than having to scroll long way down.

oscerd:
Fixes #448


oscerd:
@davsclaus the folder structure in the jar is preserved.

orpiske (approved):
I like it. Looks cleaner this way IMHO. I'll need to double check Camel JBang if nothing breaks, but I believe changes - if there are any - should be minimal.

davsclaus (approved):

PR #626, opened 2021-12-14
Rename cloudwatch cw_namespace property

tplevko:
Unify the `cw_namespace` from cloudwatch kamelet to use the camelCase. Rename to `cwNamespace`.


oscerd (approved):
Thanks a lot.

PR #671, opened 2022-01-05
Added a first version of Camel-Kamelets-Bom

oscerd:
For the moment just with commons-dbcp2 declared.

I added also a groovy script adapted for upgrading the kamelets in case a version will be changed in parent pom and in the bom.


oscerd:
Related to #670

Issue #673, opened 2022-01-05
Kamelet for ElasticSearch Sink provides no mechanism to set indexId or ce-indexId


ReggieCarey:
Greetings,

I've got  a Kamelet that sources a KafkaTopic and sinks to ElasticSearch.

The cloud event payload is JSON.

The payload does contain a unique identifier that I want to use as indexId, I cannot figure out how to use "insert-field-action" or "insert-header-action" to inject indexId/ce-indexId with a unique value from the payload (or for that matter ce-attributes).  Attempts to use "extract-field-action" might be of use but there does not appear to be any way to actually use what was extracted. (I was hoping I could extract a field and then use the contents to insert a header or insert a field).

Frankly the Kamelet documentation is useless.  It provides no data other than boilerplate as if automatically generated.  No explanation of fields, their use, or meaningful examples.  See the next three code blocks as exemplars:

```
  steps:
    - ref:
        kind: Kamelet
        apiVersion: camel.apache.org/v1alpha1
        name: insert-field-action
      properties:
        field: The Field
        value: The Value
```

or 

```
  steps:
    - ref:
        kind: Kamelet
        apiVersion: camel.apache.org/v1alpha1
        name: extract-field-action
      properties:
        field: The Field
```

or

```
  steps:
    - ref:
        kind: Kamelet
        apiVersion: camel.apache.org/v1alpha1
        name: insert-header-action
      properties:
        name: headername
        value: The Value
```

Do I have to roll my own Kamelet to do something so obviously simple?

I must be missing something.

Thanks,
Reggie

djencks:
Well, the per-kamelet docs are completely automatically generated from the yml kamelet descriptor.  I share your frustration.  Writing the latest version of the docs generation I was hoping to understand what could be customized in a kamelet and am still completely mystified.

oscerd:
The documentation needs to be improved. Contributions are always welcome. This is an open source project, anyone could help. What is unacceptable is reading something like the generated docs is useless, this is really unpleasant to read for the contributors working on this. I'll try to provide an example tomorrow or next week for this particular case.

oscerd:
David, probably you don't have an idea of how customizing a kamelet because you're only focusing on docs and website. Maybe if you ever try to run a real kamelets you will have an initial idea of what you're talking about.

djencks:
OK, so I read https://camel.apache.org/camel-k/next/kamelets/kamelets-user.html and have perhaps some clues and some questions.
[...]

PR #681, opened 2022-01-07
Extract Field Action: Make it possible to set if the extracted field should be in the body or in a particular header

### LINKED ISSUE #677: Extract Field Action: Make it possible to set if the extracted field should be in the body or in a particular header

oscerd:
Fixes #677

### DIFF
--- a/kamelets/extract-field-action.kamelet.yaml
+++ b/kamelets/extract-field-action.kamelet.yaml
@@ -29,14 +29,30 @@ metadata:
 spec:
   definition:
     title: "Extract Field Action"
-    description: "Extract a field from the body"
+    description: |-
+      Extract a field from the message body.
+
+      The extract field action expected an application/json content type. 
+
+      The field parameter allows to specify which field of the json the user wants to extract. By default the message body will be overriden with the extracted field.
+
+      The optional parameter headerOutput allows the user to specify wheter the extracted field should be stored in a message header named 'CamelKameletsExtractFieldName', leaving the message body untouched.
+
+      The headerOutput is particulary useful in case you would like to reuse an extracted field as parameters for another header, for example.
     required:
       - field
     properties:
       field:
         title: Field
-        description: The name of the field to be added
+        description: The name of the field to extract
         type: string
+      headerOutput:
+        title: Header Output
+        description: If enable the action will store the extracted field in an header named CamelKameletsExtractFieldName
+        type: boolean
+        default: false
+        x-descriptors:
+        - 'urn:alm:descriptor:com.tectonic.ui:checkbox'
     type: object
   dependencies:
   - "github:apache.camel-kamelets:camel-kamelets-utils:main-SNAPSHOT"
@@ -61,6 +77,9 @@ spec:
       - set-property:
           name: "field"
           constant: "{{field}}"
+      - set-property:
+          name: "headerOutput"
+          constant: "{{headerOutput}}"
       - bean: "org.apache.camel.kamelets.utils.transform.ExtractField"
       - choice:
           when:
--- a/library/camel-kamelets-utils/src/main/java/org/apache/camel/kamelets/utils/transform/ExtractField.java
+++ b/library/camel-kamelets-utils/src/main/java/org/apache/camel/kamelets/utils/transform/ExtractField.java
@@ -27,11 +27,16 @@ import java.util.Map;
 
 public class ExtractField {
 
-    public void process(@ExchangeProperty("field") String field, Exchange ex) {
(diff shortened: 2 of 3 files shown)


oscerd:
Related to #673

kamelets/extract-field-action.kamelet.yaml
```diff
@@ -29,14 +29,30 @@ metadata:
 spec:
   definition:
     title: "Extract Field Action"
-    description: "Extract a field from the body"
+    description: |-
+      Extract a field from the message body.
+
+      The extract field action expected an application/json content type.
```
ReggieCarey, line 35: The extract field action expects an application/json content type.

kamelets/extract-field-action.kamelet.yaml
```diff
@@ -29,14 +29,30 @@ metadata:
 spec:
   definition:
     title: "Extract Field Action"
-    description: "Extract a field from the body"
+    description: |-
+      Extract a field from the message body.
+
+      The extract field action expected an application/json content type. 
+
+      The field parameter allows to specify which field of the json the user wants to extract. By default the message body will be overriden with the extracted field.
```
ReggieCarey, line 37: The field parameter allows you to ...

library/camel-kamelets-utils/src/main/java/org/apache/camel/kamelets/utils/transform/ExtractField.java
```diff
@@ -27,11 +27,16 @@
...
+    public void process(@ExchangeProperty("field") String field, @ExchangeProperty("headerOutput") boolean headerOutput, Exchange ex) {
+        final String EXTRACTED_FIELD_HEADER = "CamelKameletsExtractFieldName";
         ObjectMapper mapper = new ObjectMapper();
         JsonNode jsonNodeBody = ex.getMessage().getBody(JsonNode.class);
         Map<Object, Object> body = mapper.convertValue(jsonNodeBody, new TypeReference<Map<Object, Object>>(){});
-        ex.getMessage().setBody(body.get(field));
+        if (!headerOutput) {
+            ex.getMessage().setBody(body.get(field));
+        } else {
+            ex.getMessage().setHeader(EXTRACTED_FIELD_HEADER, body.get(field));
```
ReggieCarey, line 38: As per my comments in the issue, maybe this is better in line 38?

ex.getMessage.setHeader(sanitized(headerOutput), body.get(field));

Where ```String sanitized(String headerOutput)``` does its best to ensure a valid headerName that does not collide with existing headers.

oscerd: Header output is a boolean. The header name is static and fixed. So, no.

oscerd:
I'm starting to be a bit bored by the way you're going through this stuff. Maybe someone else will want to help you. Have a nice weekend

library/camel-kamelets-utils/src/main/java/org/apache/camel/kamelets/utils/transform/ExtractField.java
oscerd: Also this is a bean that should be simple as simple. I troducing more logic to check stuff like headers existence is too much, there is for example the has header action or you can write your own kamelet action doing all the logic you want. But I'd probably leave the beans simplest as possible.

ReggieCarey:
I don't understand - you seem to think I'm attacking you.  Nothing could be farther from the truth.

You want community involvement but every time a question, suggestion or involvement is offered, you're offended.

If you don't like my suggestions/questions that's fine.  But I'm trying to keep this conversation from devolving into something inappropriate for public discourse on an open source project.  I truly am interested in a dialog about improving this for myself and others.

The issue tool is intended to allow public discussion and debate as well as to get ideas circulated that improves the product for all.  Some ideas are great and some not so great. Maybe my suggestions don't align with the intended product direction. 

Remember that the only people willing to spend any time at all writing issues and commenting have a vested interest in seeing this project succeed for everyone.  Attacking people that raise issues with what has been produced as open source is counter productive to the open source ideals and serves only to push people away.

If this is not the venue by which we community members make suggestions, then what is?

oscerd:
I don't discuss the where, or the issue, I'm discussing the way. But it might be it's just me.
You and anyone else are always welcome to discuss and open issues. Probably i was already upset by all the reksted discussion from yesterday and I'm still upset, but not with you in particular. 

By the way let's have another discussiin after the weekend. Thanks for the suggestions and for the help.

oscerd:
I found some time to answer now with a bit more lucidity than yesterday. I answered on the issue related to ES Index Sink Kamelet. Hope this clarifies some stuff, but lets continue the discussion there. Cheers.

PR #683, opened 2022-01-07
Insert Header Action: Use Simple as default and not constant for the value parameter binding

### LINKED ISSUE #682: Insert Header Action: Use Simple as default and not constant for the value parameter binding

oscerd:
Fixes #682


oscerd:
Related to #673

Issue #460, opened 2021-08-11
replace-field-action is not adding the renames field to enabled list by default


hguerrero:
When using the replace-field-action I would expect that the fields in the `renames` section will automatically be added to the `enabled`, and not need to add them manually.

oscerd:
What we are doing is similar to Kafka smt. Don't expect the same behavior.

oscerd:
This has been fixed with #697 because the disabled and enabled fields are now not mandatory.

PR #697, opened 2022-01-12
Replace-field-action disabled and enabled are mandatory

### LINKED ISSUE #461: replace-field-action disabled and enabled are mandatory
Currently, when using the replace-field-action the disabled and enabled fields are mandatory. If no fields are disabled, I still need to add it with an empty value

oscerd:
Fixes #461


oscerd:
Fixes #460

PR #731, opened 2022-01-24
Fix #729 to use toD for dynamic to eip

### LINKED ISSUE #729: Use toD for dynamic to in Kamelets
The correct EIP name is `toD` but we use either `to-d` or `tod` in the kamelets.

You can find them via git grep xxxx

They should be renamed to `toD`

davsclaus:


davsclaus:
Its a long time since I have helped kamelets, so let me try an easy one ;)

oscerd (approved):

PR #750, opened 2022-01-27
chore: move deprecated flow to template

squakez:
This PR span over all the list of Kamelets as we've deprecated since a while `.spec.flow` parameter in favour of `.spec.template`. https://github.com/apache/camel-k/pull/2932 will remove it from next `camel-k` spec (1.9), so we must apply this change before that PR is merged as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,7 +89,7 @@ spec:
   types:
     out:
       mediaType: text/plain
-  flow:
+  template:
     from:
       uri: timer:tick
       parameters:
@@ -128,7 +128,7 @@ spec:
   - camel:telegram
   - mvn:org.apache.commons:commons-vfs2:2.7.0
   - github:apache/camel-kamelets
-  flow:
+  template:
     # ...
 ```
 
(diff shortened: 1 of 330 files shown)


squakez:
The validation is failing #751. Will need to fix it along this PR.

djencks:
At least https://camel.apache.org/camel-k/next/kamelets/kamelets-dev.html has `flow` in the samples.  I didn't look in your other PRs, I'll plan to fix this if you haven't.

oscerd (approved):
One thing we could do is pointing the camel dependencies to 3.15.0-SNAPSHOT in library and everywhere else on main branch, otherwise this may create confusion

oscerd:
Let's merge this one for beginning.

oscerd:
Sorry I had to revert. We have multiple place where we need to address this:
- the catalog is using the camel-k extension from fabric8 kubernetes-client and there is no trace of spec.template directly there
- the validator needs to point to an updated model for camel-k crd etc. and still we cannot use it
- in this situation test on catalog will fail, validation will fail etc.
- We could leave the PR open @squakez and wait for a more stable situation for updating main here, can you place create a new PR? Thanks and sorry

squakez:
No problem. I was planning to close it anyway or rebase after #751 is fixed.

PR #754, opened 2022-01-28
chore: use template in favour of flow spec

squakez:
With this PR we're enforcing the usage of `.kamelet.spec.template` against `.kamelet.spec.flow`. The latter is deprecated since a while and will be soon removed from API specification.

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,7 +89,7 @@ spec:
   types:
     out:
       mediaType: text/plain
-  flow:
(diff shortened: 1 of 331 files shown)


oscerd (approved):

davsclaus (approved):

squakez:
Failing tests do not seems to be related to this PR.

PR #762, opened 2022-01-31
WIP: Oracle Database Sink & Source

indiealexh:
Add Oracle Database sink and source Kamelet

### DIFF
--- a/kamelets/oracle-database-sink.kamelet.yaml
+++ b/kamelets/oracle-database-sink.kamelet.yaml
@@ -56,7 +56,7 @@ spec:
         title: Server Port
         description: Server Port for the data source
         type: string
-        default: 5432
+        default: 1521
       username:
         title: Username
         description: The username to use for accessing a secured Oracle Database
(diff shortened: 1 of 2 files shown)


oscerd:
We cannot use the drive from ora le due to license. So I would remove the explicit dependency and add the instruction in the description.

indiealexh:
> We cannot use the drive from ora le due to license. So I would remove the explicit dependency and add the instruction in the description.

Can you point to an example of where this has been done before? Perhaps I should just make a generic JDBC Kamelet?

oscerd:
Hello, 

You can do something like this: https://github.com/apache/camel-kamelets/blob/main/kamelets/sqlserver-source.kamelet.yaml#L35

Personally I think makes a generic jdbc kamelet it's a bit too generic. The aim of Kamelets is being able to grab one of them to integrate with a specific external service or db. So IMO, it would be really better to have one Kamelet for each Database kind.

oscerd:
Btw the skeleton of the Kamelet is good, you just need to remove the dependency and add the instruction to the description. Thanks a lot for working on this.

oscerd:
Did you have time to have a look at updating this? Thanks

oscerd:
If you don't mind, I'll clean up and merge this PR.

indiealexh:
> If you don't mind, I'll clean up and merge this PR.

Yeah, sorry been crazy busy at work and havn't had a chance to follow up with this.

PR #763, opened 2022-02-01
Change to flow description

squakez:
Better usage of `template` in the description

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,15 +102,15 @@ spec:
 ```
 
 
-### Flow Code
+### Template Code
 
-The Camel route that defines the behavior of the Kamelet **MUST** be provided in YAML flow syntax.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -102,15 +102,15 @@ spec:
 ```
 
 
-### Flow Code
+### Template Code
 
-The Camel route that defines the behavior of the Kamelet **MUST** be provided in YAML flow syntax.
+The Camel route that defines the behavior of the Kamelet **MUST** be provided in YAML flow syntax (in `.spec.template` parameter).
```
djencks, line 107: When I google "YAML flow syntax" all the hits are about embedding json in yaml (i.e. curly brackets rather than  indentation).  None of the kamelets use format in the flow/tempate section, so I think you need to provide a link to a description of what you mean.

oscerd (approved):
We can add more at later stage

PR #834, opened 2022-03-16
chore: Add Kamelet rest-openapi-sink

christophd:
New Kamelet loads OpenAPI specification from (http/https URL) and calls an operation on the HTTP service. The request is generated respecting the rules given in the OpenAPI specification (e.g. path parameters, Content-Type).


test/rest-openapi-sink/rest-openapi-sink-binding.yaml
```diff
@@ -0,0 +1,29 @@
+apiVersion: camel.apache.org/v1alpha1
```
oscerd, line 1: Can you add license?

test/rest-openapi-sink/openapi.json
```diff
@@ -0,0 +1,289 @@
+{
```
oscerd, line 1: Can you add license?

test/rest-openapi-sink/rest-openapi-sink.feature
```diff
@@ -0,0 +1,52 @@
+Feature: REST OpenAPI Kamelet sink
```
oscerd, line 1: Can you add license?

kamelets/rest-openapi-sink.kamelet.yaml
```diff
@@ -0,0 +1,57 @@
...
+spec:
+  definition:
+    title: REST OpenAPI Sink
+    description: Load OpenAPI specification from URI and call an operation on the HTTP service. The request is generated respecting the rules given in the OpenAPI specification (e.g. path parameters, Content-Type).
+    required:
+    - specification
+    - operation
+    type: object
+    properties:
+      specification:
```
oscerd, line 38: This is good if we run in camel-jbang or camel-kamelet-main, but if we have to deal with camel-k it might be a problem to point a physical file. I mean we should have a volume mounted and containing the specification. In some of the transformation actions, like the templating ones, we are passing a base64 encoded string of this kind of parameters. I think we should harmonize the behavior, but probably we need to change the templating ones and this is good. @davsclaus @lburgazzoli what do you think?

oscerd (approved):
LGTM, just minor findings.

kamelets/rest-openapi-sink.kamelet.yaml
davsclaus: I assume you can use file: or classpath: to refer to a local spec file. And if you use camel-k in container, then you would need to include this file in the built image.

Maybe using runtime resources
https://camel.apache.org/camel-k/1.8.x/configuration/runtime-resources.html

christophd: Yes, the rest-openapi Camel component should support both remote and local resources already.

So, I think runtime resources is the way to go as the Camel K operator takes care on making these resources available as local resources in the integration container. Would have to check weather KameletBinding already supports these kind of resources though

test/rest-openapi-sink/openapi.json
christophd: Standard Json does not support comments so it is hard to add the license here. Have added the license headers to other resources though. Thanks for spotting

kamelets/rest-openapi-sink.kamelet.yaml
oscerd: Sounds good. So for 0.8.x, I'll change this stuff for the template actions. @lburgazzoli sounds good for you too?

oscerd (approved):

oscerd:
Thanks @christophd

PR #846, opened 2022-03-21
chore: Overwrite JitPack dependencies for pull request builds

christophd:
This change in GitHub workflow definition enables PRs to use new/changed utility classes in module camel-kamelets-utils. The workflow overwrites the JitPack dependency coordinates in all Kamelets to point to the actual PR head ref instead of base (e.g. main-SNAPSHOT).

This ensures that the automated tests use the utility classes from that specific PR instead of those from the base branch (e.g. main-SNAPSHOT). So pull request builds are able to have a green GitHub CI workflow even if the base branch is not aware of the utility class before the merge.

### DIFF
--- a/.github/workflows/yaks-tests.yaml
+++ b/.github/workflows/yaks-tests.yaml
@@ -54,6 +54,17 @@ jobs:
     steps:
     - name: Checkout code
       uses: actions/checkout@v2
+    - name: Set JitPack coordinates for pull requests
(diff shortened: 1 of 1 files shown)


oscerd (approved):

christophd:
please bear with me as I need to find the proper github base ref value to extract from github context in order to make the PR a success

christophd:
@oscerd ready now, it works!

Issue #843, opened 2022-03-18
Add Google BigQuery sink kamelet


claudio4j:

claudio4j:
I am working on this

PR #849, opened 2022-03-22
Add Google Big Query Sink Kamelet

claudio4j:
https://github.com/apache/camel-kamelets/issues/843

### DIFF
--- a/docs/modules/ROOT/nav.adoc
+++ b/docs/modules/ROOT/nav.adoc
@@ -66,6 +66,7 @@
 * xref:github-pullrequest-comment-source.adoc[]
 * xref:github-pullrequest-source.adoc[]
 * xref:github-tag-source.adoc[]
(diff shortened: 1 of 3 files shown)


oscerd (approved):

oscerd (changes requested):
Looking a bit better at this one, the payload in the body should be a map or a list. So I'm not sure this will work out of the box, since sources are not mapping lists or maps directly.

claudio4j:
The kamelet can expect a json as body and I can add a step to transform the json to map or list of map

oscerd:
This needs to be documented in the description, because otherwise a user might not know what he has to do.

oscerd (approved):

oscerd (changes requested):
Please add 

```
  types:
    in:
      mediaType: application/json
```

oscerd (approved):

PR #862, opened 2022-03-28
feature(salesforce-sink): Add, update, and delete salesforce sink

Delawen:


Delawen:
Pending: add secrets to Github and make the secrets load properly on the tests.

Delawen:
@christophd can you help me with the secrets?

oscerd (approved):

kamelets/salesforce-delete-sink.kamelet.yaml
```diff
@@ -0,0 +1,107 @@
...
+      - userName
+      - password
+    type: object
+    properties:
+      sObjectName:
+        title: Object Name
+        description: Type of the object
+        type: string
+        example: Contact
+      sObjectId:
```
lburgazzoli, line 47: I wonder if having to set the `sObjectId` as a parameter is right because that would mean that there is a 1-1 relation between the Kamelet and the object to delete (unless a `toD` is used but then it would make the integration a template of templates)

kamelets/salesforce-delete-sink.kamelet.yaml
```diff
@@ -0,0 +1,107 @@
...
+        title: Password
+        description: The Salesforce user password
+        type: string
+        format: password
+        x-descriptors:
+        - urn:alm:descriptor:com.tectonic.ui:password
+        - urn:camel:group:credentials
+  types:
+    in:
+      mediaType: application/json
```
lburgazzoli, line 86: What does the kamelet expect as an input ? 
Looking at the kamelet definition, it looks like everything that is needed can be expressed as parameter.

kamelets/salesforce-delete-sink.kamelet.yaml
Delawen: So if it is not sent as parameter... extract it from the body?

kamelets/salesforce-delete-sink.kamelet.yaml
Delawen: Yes, there is no need for body here. Just the id. So do I change it to send the id as a string in the body and that's it?

kamelets/salesforce-delete-sink.kamelet.yaml
lburgazzoli: That may be an option
I don't know if you want to make both the `sObjectType` and `sObjectId` something you expect in the body

Delawen: That's also an option. Let me try :)

Delawen:
Disabled tests directly to salesforce api to prevent problems if the api is down.

kamelets/salesforce-delete-sink.kamelet.yaml
Delawen: Improved the documentation of what kamelets expect as input.

Delawen: Added examples. The schema (except on the delete sink) is dependent on the salesforce object type.

lburgazzoli: In this case, it does not seems that the scheme is dependant on the object type as it is always 

```json
{ "sObjectId": "XXXXX0", "sObjectName": "Contact" }
```

Unless I missed how this kamelet works

Delawen: That's the delete one, you must specify the object name (table) and object id (key).

People that understand salesforce should know what that means...

Delawen: Where do you specify the schema? On the 
types:
  in:
??

kamelets/salesforce-delete-sink.kamelet.yaml
```diff
@@ -0,0 +1,103 @@
...
+          userName: "{{userName}}"
+          password: "{{password}}"
+          loginUrl: "{{loginUrl}}"
+    from:
+      uri: kamelet:source
+      steps:
+        - set-header:
+            name: sObjectId
+            jsonpath: "$['sObjectId']"
+        - set-header:
```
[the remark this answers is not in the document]
Delawen: Done.

kamelets/salesforce-update-sink.kamelet.yaml
```diff
@@ -0,0 +1,111 @@
...
+      - userName
+      - password
+    type: object
+    properties:
+      sObjectName:
+        title: Object Name
+        description: Type of the object
+        type: string
+        example: Contact
+      sObjectId:
```
lburgazzoli, line 51: Is it expected that the user should provide the sObjectId as part of the properties ? 
it seems that this could be very limited as what if you want to update another object ?

kamelets/salesforce-delete-sink.kamelet.yaml
```diff
@@ -0,0 +1,116 @@
...
+        format: password
+        x-descriptors:
+        - urn:alm:descriptor:com.tectonic.ui:password
+        - urn:camel:group:credentials
+  types:
+    in:
+      mediaType: application/json
+      schema:
+        type: object
+        properties:
```
lburgazzoli, line 82: if any of the property is mandatory, it should be listed among the required fields

kamelets/salesforce-update-sink.kamelet.yaml
Delawen: If they provide the full salesforce object, then that contains the sObjectId.

If they don't provide the full salesforce object, then it must be defined outside the body. That's why sObjectId is not required.

kamelets/salesforce-delete-sink.kamelet.yaml
Delawen: Done.

kamelets/salesforce-delete-sink.kamelet.yaml
Delawen: Done.

kamelets/salesforce-update-sink.kamelet.yaml
lburgazzoli: > That's why sObjectId is not required.

That's partially true as it is not marked as required parameter but it is not optional in the route so the route is likely failing in case the id is not set.

I would also argue that we should be a little opinionated when making a kamelet as if the end result is about wrapping an endpoint, then there's a very little advantage in using a kamelet vs using the component. Having a non clear interface also make it a little bit more complex to use a Kamelet from tooling POV (Kaoto, Karavan, OpenShifr Dev Console, etc).

That said, I'm fine to keep `sObjectName` and `sObjectId` if needed but let's try to have a concrete use case for when setting the `sObjectId` as part of a kamelet configuration would be required.

Delawen: > That's partially true as it is not marked as required parameter but it is not optional in the route so the route is likely failing in case the id is not set. 

Oh, you are right.

I think using salesforce objects is a more complex usecase than just sending the key-value pairs for attributes changing.

But maybe for real users that use salesforce I am saying a heresy :D

Delawen: I made them truly optional, I don't feel I know salesforce enough to be opinionated :)

lburgazzoli: > I think using salesforce objects is a more complex usecase than just sending the key-value pairs for attributes changing.
> 
> But maybe for real users that use salesforce I am saying a heresy :D

I don't know either and that's why I personal would start with something opinionated and I would iterate over it once we have clarity. Adding an optional field is much easier an safe than removing one as that would break the contract.

Delawen: Then I will work only with key-value pairs and let's see if someone complains :)

Delawen: Pushed.

Delawen:
The failed test is not related to this PR...

Delawen:
Should I maybe rebase? Or is it flaky tests and will be the same?

oscerd:
It's not failing on other PRs.

oscerd:
Let's try to rebase.

christophd:
after the rebase the test is green, YAY!

this Kafka test has been flaky before (~50% failing) but I think I have fixed some issues with it. as @oscerd has mentioned the test is stable on other PRs. So lets's keep an eye on that one as it may still be flaky (maybe ~10%).

PR #886, opened 2022-04-09
Issue 885

agiertli:


agiertli:
Hi @oscerd 

We are using this kamelet with one extra dependency:

    - 'mvn:org.apache.activemq:activemq-http:5.16.2'

This allows communication with AMQ Broker over HTTP protocol (as opposed to tcp for example). Since it hardcodes the library version I was hesitant whether this should be included as well and not sure how widely popular HTTP integration is.

oscerd (approved):

oscerd:
No problem for the dependency, we could also externalize the version. I'll do that.

oscerd:
Thanks.

Issue #866, opened 2022-03-28
override example template for these action kamelets


MelissaFlinn:
By default, the Kamelets template auto generates examples. We discovered that the auto-generated examples for the following Kamelets do not work. (As a workaround, we provided working examples in the Kamelets Reference document.)
avro-deserialize-action-binding.yaml
avro-serialize-action.kamelet.yaml
has-header-filter-action.kamelet.yaml
insert-action.kamelet.yaml
protobuf-deserialize-action.kamelet.yaml
protobuf-deserialize-action.kamelet.yaml

oscerd:
This is a description not really complete. What is wrong exactly?

squakez:
If I understand correctly, the template we use to render Kamelets documentation is not correct for action kind Kamelets.

oscerd:
The reported list is just a subset, so only this subset has something wrong, but what is it? Waiting for clarification.

MelissaFlinn:
What is wrong: the auto-generated examples for those Kamelets do not work.

MelissaFlinn:
The examples need additional steps - such as adding a serialize/deserialize step. The way that the template generates examples now, we cannot override the generated ones to provide examples that are correct/useful.

oscerd:
Now I have a description.

claudio4j:
I am investigating this issue. Looking at the [avro-deserialize-action](https://camel.apache.org/camel-kamelets/next/avro-deserialize-action.html) and the corresponding binding example file [avro-deserialize-action-binding.yaml](https://github.com/apache/camel-kamelets/blob/0.7.x/templates/bindings/camel-k/avro-deserialize-action-binding.yaml)

It fails with `JsonParseException: Needed to read 36 bytes, missed 36 before end-of-input`

The kamelet binding should be fixed by having a json message in the timer and additional steps to convert string -> json -> avro.

I had not run all the other auto generated kamelet biding to see if they work. 

To fix this, I propose one of the following options:
1. Have all the kamelet bindings in `templates/bindings/camel-k` directory in working condition and have `docs/modules/ROOT/examples/js/kamelets.js` to source these files as examples instead to auto generate them. And have only one kafka or knative channel sink in the example binding file.

2. Fix only the non working binding example files in `templates/bindings/camel-k` directory, add a comment marker to it, and change `docs/modules/ROOT/examples/js/kamelets.js` to source this example file into the generated html doc, instead to auto generated the binding example.

PR #897, opened 2022-04-25
Update non-working examples in Kamelet Catalog docs

claudio4j:
Fix https://github.com/apache/camel-kamelets/issues/866

* There are some generated kamelet bindings are invalid, the auto
  generation cannot resolve all the required steps, so there is
  a need to create it manually
* Add a comment marker in kamelet binding examples for the auto
  generation to skip its creation
* The doc generation source the kamelet binding example file when it
  doesn't auto generate the example
* the camel-kamelets directory is hardcoded in kamelets.js as it is
  assumed this generation mechanism already assumes this directory

### DIFF
--- a/docs/modules/ROOT/examples/js/kamelets.js
+++ b/docs/modules/ROOT/examples/js/kamelets.js
@@ -16,6 +16,8 @@
  */
 
 const util = require('camel-website-util')
+const fs = require('fs')
+const yaml = require('js-yaml');
 
 const QUOTED_CHARS = /[$`"\\]/g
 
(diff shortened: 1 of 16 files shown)


claudio4j:
I would like a review about this approach, if everything is ok, I can work to fix the remaining non working kamelet binding examples.

oscerd (approved):

claudio4j:
@oscerd @squakez this is ready for review.

oscerd (approved):

oscerd:
Thanks

docs/modules/ROOT/examples/js/kamelets.js
```diff
@@ -16,6 +16,8 @@
  */
 
 const util = require('camel-website-util')
+const fs = require('fs')
+const yaml = require('js-yaml');
```
zregvart, line 20: If you're importing here you also need to adjust package.json in camel-website, i.e. add `js-yaml` to `devDependencies`. You also need to make sure that the `js-yaml` is placed in the PnP cache. Or avoid using `js-yaml` or any non-managed dependency.

I can see the net positive of having them versioned within the module. But I don't see a way for them to be tested, debugged or linted here. So until this is figured out I can't recommend doing this here.

oscerd: Do we have to revert? The website build is stuck

claudio4j: Ops, I will push the js-yaml dependency change to camel-website. Just forgot it.

claudio4j:
https://github.com/apache/camel-website/pull/835

PR #919, opened 2022-05-06
Added default value for keyName in aws-s3-sink.kamelet.yaml

valdar:


kamelets/aws-s3-sink.kamelet.yaml
```diff
@@ -121,6 +121,12 @@ spec:
             - set-header:
                 name: CamelAwsS3Key
                 simple: "${header[ce-file]}"
+          - simple: "${properties:keyName:null} == 'null'"
+            steps:
+            - set-header:
+                name: CamelAwsS3Key
+                simple: "${exchangeId}"      
+                #
```
oscerd, line 129: I don"t think this line is needed

oscerd (changes requested):
Minor one

kamelets/aws-s3-sink.kamelet.yaml
valdar: I have added a comment there that was lost in the first commit.

oscerd (approved):

PR #982, opened 2022-07-25
Adding a new kamelet that uses the nominatim geocoder

Delawen:
I created this kamelet because I want to use it on a demo next month. It has the geocoder camel component behind.

Given a URL of the geocoder server (Nominatim is FOSS and you can either deploy your own or use an existing one) as header and a text in the body, it searches for the geographic coordinates and the details of the address passed in the body.


oscerd (changes requested):
Please add the type.

Delawen:
You mean the output type? I am not sure why the validation is failing, I am defining the objects there.

oscerd:
You need to add the type: object to the definition like https://github.com/apache/camel-kamelets/blob/main/kamelets/aws-s3-sink.kamelet.yaml#L46

Delawen:
Is it not here? https://github.com/apache/camel-kamelets/pull/982/files#diff-267c7bda93e779595b51f4629a85908f4b748af2bb09d97aeaf3477614b92022R59

Or is `types` deprecated in favor of singular `type`?

Delawen:
I'm trying to mimic https://github.com/apache/camel-kamelets/blob/main/kamelets/salesforce-delete-sink.kamelet.yaml#L77-L87 but maybe I can't have an object inside an object?

oscerd:
You need to add type: object in the kamelet definition. Types is a different thing as far as I know.

Delawen:
I was too smart then. Let's see if that works

oscerd (changes requested):
The indentation is wrong, type:object should be aligned with required.

Delawen:
Something else is missing, testing the validation locally now... (which I should have started with)

oscerd (approved):

Issue #980, opened 2022-07-22
Reopen of #962 - ISSUE WITH KAMELET azure-servicebus-source output is message length instead of message


bjethwan:
ISSUE WITH KAMELET azure-servicebus-source output is message length instead of message

I had below KameletBinding working with Timer source but with Azure ServiceBus.
I was getting a type conversion error
Caused by: org.apache.camel.NoTypeConversionAvailableException: No type converter available to convert from type: com.azure.core.util.BinaryData to the required type: java.io.InputStream

Hence, I added a JSON Serializer in between

```
apiVersion: camel.apache.org/v1alpha1
kind: KameletBinding
metadata:
  name: aws-s3-streaming-upload-sink-binding
spec:
  source:
    ref:
      kind: Kamelet
      apiVersion: camel.apache.org/v1alpha1
      name: azure-servicebus-source
    properties:
      connectionString: Endpoint=<<>>/;SharedAccessKeyName=RootManageSharedAccessKey;SharedAccessKey=<<>>
      topicOrQueueName: <<>>
      serviceBusReceiveMode: RECEIVE_AND_DELETE
  steps:
  - ref:
      kind: Kamelet
      apiVersion: camel.apache.org/v1alpha1
      name: json-serialize-action
  sink:
    ref:
      kind: Kamelet
      apiVersion: camel.apache.org/v1alpha1
      name: aws-s3-streaming-upload-sink
    properties:
      bucketNameOrArn: arn:aws:s3:::<<>>
      keyName: timer-source
      region: ap-south-1
      accessKey: <<>>
      secretKey: <<>>
      batchMessageNumber: 5
```

Now the PROBLEM is that instead of capturing data it's capturing length of the message
{"length":54}{"length":54}{"length":54}{"length":54}{"length":54}

How to fix this?
Do I need to configure the azure-servicebus-source Kamelet to make it output String/JSON? Is it reasonable to configure the Kamelet or should I copy and configure that one?

bjethwan:
@oscerd 
Hi Andrea - I have reopened the issue.
Unless I update the existing azure-servicebus-source to emit string messages.
KameletBinding doesn't provide a way to correctly convert the Byte stream to String.
As you can see that when I try to use json-serialize-action, instead of the actual message I am getting message length.

oscerd:
The message from servicebus is always binary. 

[...]

PR #983, opened 2022-07-25
ISSUE WITH KAMELET azure-servicebus-source output is message length i…

oscerd:
…nstead of message


oscerd:
Fixes #980

oscerd:
@bjethwan

With this you'll get the bytes related to the binaryData.

oscerd:
So you can serialize as json or you can convert the payload to a String.

bjethwan:
@oscerd Thank you !
I was using the same in my local test environment to make the azure kamelet work.

Thank you so much.

PR #988, opened 2022-07-27
New action that gets features from OGC API server.

Delawen:
See https://www.ogc.org/standards/ogcapi-features

You can either call for a full geoJSON with all the results or iterate over each result using the `split` parameter.

This is a new geospatial API standard that replaces WFS servers.

Kamelet Binding used to test:

```
apiVersion: camel.apache.org/v1alpha1
kind: KameletBinding
metadata:
  name: integration
spec:
  source:
    ref:
      apiVersion: camel.apache.org/v1alpha1
      name: timer-source
      kind: Kamelet
    properties:
      period: '5000000'
      message: 'NONLD_AREA=0.0&NAME=Croydon'
  steps:
  - ref:
      apiVersion: camel.apache.org/v1alpha1
      name: ogcapi-features-action
      kind: Kamelet
    properties:
      collection: activity_level_ldn
      url: https://emotional.byteroad.net/
  sink:
    ref:
      apiVersion: camel.apache.org/v1alpha1
      name: log-sink
      kind: Kamelet
    properties:
      showAll: true

```

### DIFF
--- a/kamelets/ogcapi-features-action.kamelet.yaml
+++ b/kamelets/ogcapi-features-action.kamelet.yaml
@@ -33,6 +33,8 @@ spec:
       Returns the items of the collection provided of an OGC API Features server. 
       The collection must be a valid collection name on the server.
       
+      Query can be defined in the body too.
+      
       See https://www.ogc.org/standards/ogcapi-features
     required:
       - url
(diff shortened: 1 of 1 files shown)


Delawen:
This is not a source because there is no default "timestamp" parameter from which to query only the last results.

oscerd (approved):

PR #999, opened 2022-08-09
feat: enable raw payload on salesforce create/update kamelets

lburgazzoli:


oscerd (approved):

oscerd (commented):
Maybe it makes sense to make it configurable with true/false flag, instead of defaulting to true.

PR #1057, opened 2022-09-30
Elasticsearch-search-source won't let me define basic authentication

### LINKED ISSUE #1056: elasticsearch-search-source won't let me define basic authentication
While trying to set the basic authentication (user, password) I got the error `Unknown parameters=[{password=xxxxxx, user=xxxxxx}]` I think the bug is really on [Camel repository](https://github.com/apache/camel) but I can't open an issue there.

The reason I think it's a bug there is that the [documentation](https://camel.netlify.app/components/latest/elasticsearch-rest-component.html) says that `user` and `password` are options but when I try to define a route directly (without using the kamelet), `elasticsearch-rest://clusterName?password=xxxxxx&user=xxxxxx` gives me the same error.

I've tried to debug the problem and notice that on `ElasticsearchEndpointConfigurer.java` the `user` and `password` are not present on the switches cases, all other options are, but maybe this is on purpose since these are the only ones with `@Metadata(secret = true)` on `ElasticsearchComponent.java`.
Also, if `ElasticsearchEndpoint` class `@UriEndpoint` annotation had the flag `lenientProperties = true`, it would skip the validation where the error originates, but dunno if it would then configure the `user` and `password` options with the correct value.

Or maybe it's just me that is missing something and can't get this to work on [camel-quarkus](https://github.com/apache/camel-quarkus).

oscerd:
Fixes #1056

### DIFF
--- a/kamelets/elasticsearch-search-source.kamelet.yaml
+++ b/kamelets/elasticsearch-search-source.kamelet.yaml
@@ -93,6 +93,9 @@ spec:
     beans:
       - name: local-es
         type: "#class:org.apache.camel.component.elasticsearch.ElasticsearchComponent"
+        properties:
+            user: "{{?user}}"
+            password: "{{?password}}"
     from:
       uri: timer:tick
       parameters:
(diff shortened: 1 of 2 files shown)


oscerd:
@MarcoMartins86

PR #1142, opened 2022-11-08
Add level property to Log Sink

tadayosi:


oscerd (approved):
Can we backport to 0.9

PR #1181, opened 2022-11-24
Fix mail-sink Kamelet

christophd:
- In fact mail-sink only supports smtp protocol. Others like pop3 and imap do not make sense for sending mails
- SMTPS protocol is also not supported because SSL context is missing
- to:{{protocol}} stopped working with latest versions so remove protocol property and always use "smtp"
- Also fix YAKS test

### DIFF
--- a/kamelets/mail-sink.kamelet.yaml
+++ b/kamelets/mail-sink.kamelet.yaml
@@ -36,11 +36,6 @@ spec:
       - password
     type: object
     properties:
-      protocol:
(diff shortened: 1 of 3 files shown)


oscerd (approved):

oscerd:
Good to backport on 0.9.x. Thanks

christophd:
@oscerd thanks! could you please merge? I need this for another PR

PR #1243, opened 2023-01-19
isolate the verifyHeader method in testSupportedHeaders to allow customizations

claudio4j:


oscerd (approved):
Thanks, this should be on 3.20.x too, I'll backport.

PR #1318, opened 2023-02-28
Issue-981: Restore optional usage of chatId Configuration Option and correct narrative

manstis:
Following discussions with @valdar it was felt that including a default value and improving the documentation was the best approach. Keeping the parameter as `required` forces Users to explicitly consider how/where to set the value. Reverting optional behaviour can lead to implicit failures that can be more difficult to diagnose.

@valdar @oscerd WDYT?

### DIFF
--- a/kamelets/telegram-sink.kamelet.yaml
+++ b/kamelets/telegram-sink.kamelet.yaml
@@ -47,7 +47,6 @@ spec:
       - `chat-id` / `ce-chatid`: overrides the default chat where messages are sent to
     required:
       - authorizationToken
-      - chatId
     type: object
     properties:
       authorizationToken:
@@ -60,7 +59,10 @@ spec:
         - urn:camel:group:credentials
       chatId:
         title: Chat ID
-        description: The Chat ID to where you want to send messages by default. Optional if the `chat-id` / `ce-chatid` header is used.
+        description: |-
+          The Chat ID to where you want to send messages by default. 
+          
+          Whilst the Chat ID is not a required Configuration Option it must be provided for every message; either as a Configuration Option or a `chat-id` / `ce-chatid` header.
         type: string
   types:
     out:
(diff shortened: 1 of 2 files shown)


kamelets/telegram-sink.kamelet.yaml
```diff
@@ -60,8 +60,14 @@ spec:
...
         title: Chat ID
-        description: The Chat ID to where you want to send messages by default. Optional if the `chat-id` / `ce-chatid` header is used.
+        description: |-
+          The Chat ID to where you want to send messages by default. 
+          
+          The Chat ID must be provided for every message; either as a Configuration Option or a `chat-id` / `ce-chatid` header.
+          
+          If the Chat ID is not explicitly provided the default value is used which may lead to delivery failures.
         type: string
+        default: dummy
```
oscerd, line 70: If you set a default value, it will be set on the endpoint by default.

oscerd (changes requested):
I don't think default should be set

kamelets/telegram-sink.kamelet.yaml
manstis: This is @valdar's preference. 

However I see your point too. Damn. Why do I always see both sides to things!?!?

The User is forced to either provide a _real_ value for the endpoint configuration or set its value in the header where it then takes precedence over the endpoint configuration. The default value is explicitly mentioned in the updated documentation and advises it _could_ lead to delivery failure.

It's a difficult situation: The parameter is _required_, it is not optional; but _where_ it can be set (endpoint or per message) means Camel fails to deploy the Kamelet if the User does not set a value in the Configuration Options when marked as `required`. 

We want to somehow impose configuration of a `chatId` on Users.

I'll chat to @valdar again.. the imposition is something we want in RHOC but seems contrary to both the underlying Camel Telegram Component and your opinion.

manstis: OK. I've restored the `required` nature of the `chatId` parameter and hopefully made the documentation bullet proof!

manstis:
Assuming this is acceptable; please confirm you'll want the other affected branches (`3.x` and `3.20.x`) corrected too.

oscerd (approved):

oscerd:
This should also be backported on 3.x and 3.20.x

manstis:
> This should also be backported on 3.x and 3.20.x

See:
- https://github.com/apache/camel-kamelets/pull/1319
- https://github.com/apache/camel-kamelets/pull/1320

PR #1328, opened 2023-03-01
chore: Add data-type-action Kamelet

christophd:
- Adds new action Kamelet to leverage Kamelet input/ouput data types
- Enables users to apply data type conversion as part of a KameletBinding
- Introduces new data type implementation representing the CloudEvents Http binding
- Add some YAKS tests to verify data-type-action Kamelet
- Use Kamelets from local directory when running YAKS tests in GitHub workflow


oscerd (approved):
This is really important: we could avoid modifying the current kamelets code and leave the choice oto the user to use or not use the action.
I think the next step should be removing the experimental kamelets then.

After that we should see if we might need to create more beans for all the supported kamelets to use in this action. But I need to dig a bit more 

Does this sounds correct to you?

christophd:
@oscerd Yes, sounds good, many thanks. I will prepare a PR to remove the experimental Kamelets as a next step

christophd:
still not sure about the name 'data-type-action' though. Something like 'apply-data-type-action' would be a bit long. any ideas?

davsclaus (approved):

oscerd:
> still not sure about the name 'data-type-action' though. Something like 'apply-data-type-action' would be a bit long. any ideas?

maybe data-type-transform-action?

PR #1329, opened 2023-03-02
Slack Sink: iconEmoji, iconUrl, username are deprecated in camel

manstis:
Please advise if you require Camel issues lodged for PRs vs third party tracking references.

The underlying Slack Component has deprecated a few [parameters](https://camel.apache.org/components/3.20.x/slack-component.html#_path_parameters_1_parameters), namely `iconEmoji`, `iconUrl` and `username`.

This PR attempts to reflect the same in the corresponding Kamelet definition.

I'm uncertain if Kamelets even support the definition of deprecated parameters. [Documentation](https://camel.apache.org/camel-k/1.11.x/kamelets/kamelets-dev.html#_step_6_describe_the_parameters) isn't clear.

I'm even less clear, to be honest, whether the `parameters` in a Kamelet map to "Components Options" or "Query parameters" or both due to some clever name-based mapping in Camel. The three affected Kamelet parameters appear as `parameters` in the Kamelet but as "Query parameters" in the Component. The Kamelet parameter `webhookUrl` relates to a [Component Option](https://camel.apache.org/components/3.20.x/slack-component.html#_component_option_webhookUrl) and [Query parameter](https://camel.apache.org/components/3.20.x/slack-component.html#_endpoint_query_option_webhookUrl).

I suspect they only map to Component Options and hence the three "Query" parameters can be removed entirely and not just deprecated.

Advice welcome.

### DIFF
--- a/kamelets/slack-sink.kamelet.yaml
+++ b/kamelets/slack-sink.kamelet.yaml
@@ -54,14 +54,17 @@ spec:
         title: Icon Emoji
         description: Use a Slack emoji as an avatar.
         type: string
+        deprecated: true
       iconUrl:
         title: Icon URL
         description: The avatar to use when sending a message to a channel or user.
         type: string
+        deprecated: true
       username:
         title: Username
         description: The username for the bot when it sends messages to a channel or user.
         type: string
+        deprecated: true
   types:
     out:
       mediaType: application/json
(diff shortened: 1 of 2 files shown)


oscerd:
Please don't use downstream JIRA. For the upstream there is no point in linking them. This is the upstream and we should focus on upstream only. I edited the text

manstis:
Closing in lieu of https://github.com/apache/camel-kamelets/issues/1330

Whatever solution is agreed will need to be enacted for this Slack Sink.

PR #1378, opened 2023-03-16
JMS AMQP 1.0 Source/Sink Kamelets: Replace QPID client with camel-amqp

gansheer:
## Motivation

The client qpid dependency does not works in native mode (stacktrace log `Caused by: java.lang.ClassNotFoundException: org.apache.qpid.jms.JmsConnectionFactory`).

## Description

The camel-amqp component is designed for the kamelets use case : supporting the AMQP 1.0 protocol through the JMS QPID Client. So replacing the client dependency by the component makes sense.


oscerd (approved):

gansheer:
Additional note : I don't need the camel-jms component since it is a dependency from camel-amqp. When removing it the validate script fails because it is not present anymore in dependencies. 
Should I remove it and make the change to adapt the kamelets (I am guessing a change of names and the apache group in metadatas at least) ?

oscerd:
I think the validation must be reviewed at all also I do think we don't think the qpid-jms-client in the bom, since it's not managed anymore now.

gansheer:
> I think the validation must be reviewed at all also I do think we don't think the qpid-jms-client in the bom, since it's not managed anymore now.

Ha you are right, I will remove the QPID client everywhere it needs to be.

zhfeng:
@gansheer I'm just curious that
> The client qpid dependency does not works in native mode

Is it Quarkus native mode? If it is, I think we need to use `org.amqphub.quarkus:quarkus-qpid-jms`. see https://github.com/amqphub/quarkus-qpid-jms

oscerd:
> @gansheer I'm just curious that
> 
> > The client qpid dependency does not works in native mode
> 
> Is it Quarkus native mode? If it is, I think we need to use `org.amqphub.quarkus:quarkus-qpid-jms`. see https://github.com/amqphub/quarkus-qpid-jms

Kamelets are not designed to work only with quarkus as runtime, it was like that in the beginning, so I don't think it's the case to add a quarkus dependency directly in the yaml.

gansheer:
@zhfeng I am talking about the library "mvn:org.apache.qpid:qpid-jms-client:1.0.0". 
I tested the modified kamelets with camel-k, which means camel-quarkus, and it works well. I don't think I can add a quarkus dependency in the Kamelets :thinking: .

zhfeng:
I get it and thanks for your clarifying.

PR #1474, opened 2023-05-18
added sslCipherSuite property so TLS can be enabled

vedran-kolka-syntio:
This PR adds a sslCipherSuite property to the jms-ibm-mq sink and source kamelets that enables using TLS with IBM MQ.


oscerd (approved):
Thanks for this

davsclaus (approved):

PR #1610, opened 2023-09-04
chore(docs): escape variable to prevent site failure

squakez:
The problem is described in #1609. Likely this one solves, but we need to verify if the added character will render in the visualization as well. In such case, we need to change the template with a replacement as suggested in #1609.


oscerd (approved):

oscerd:
It seems it's not working.

oscerd:
For the moment I would remove the example line and maybe add an example in the description.

squakez:
> For the moment I would remove the example line and maybe add an example in the description.

Yes. I'll go on this direction and will create a follow up issue.

PR #1612, opened 2023-09-05
Create salesforce-composite-upsert-sink.kamelet.yaml

Ziauddin135:
This is the sink Kamelet which allows to facilitate Upsert the composite list of sObjects in the Salesforce.


oscerd (approved):

oscerd:
Thanks

PR #1638, opened 2023-09-15
Enhance application-x-struct data type

christophd:
- Support empty message body value and initialize the application-x-struct data type with empty JsonNode {} value
- Allows follow-up action Kamelets such as insert-field-action to add more fields to the empty JsonNode body

Example:

1. webhook-source creates empty message body
2. application-x-struct data-type used as output data type initializes empty JsonNode
3. insert-field action Kamelet uses header fields to set fields on the JsonNode
4. Json struct gets serialized to String value as outcome

### DIFF
--- a/library/camel-kamelets-utils/src/main/java/org/apache/camel/kamelets/utils/format/converter/json/JsonStructDataType.java
+++ b/library/camel-kamelets-utils/src/main/java/org/apache/camel/kamelets/utils/format/converter/json/JsonStructDataType.java
@@ -20,6 +20,7 @@ package org.apache.camel.kamelets.utils.format.converter.json;
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.io.InputStream;
+import java.nio.charset.StandardCharsets;
 
 import com.fasterxml.jackson.databind.JsonNode;
 import org.apache.camel.CamelExecutionException;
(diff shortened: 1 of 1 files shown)


christophd:
Webhook called with `http://localhost:8080/bookings?name=foo&amount=100`
result in Json `{"name": "foo", "amount": "100"}` on Kafka sink

```yaml
apiVersion: camel.apache.org/v1alpha1
kind: KameletBinding
metadata:
  name: webhook-source-booking-pipe
spec:
  source:
    ref:
      kind: Kamelet
      apiVersion: camel.apache.org/v1
      name: webhook-source
    properties:
      subpath: bookings
    dataTypes:
      out:
        format: application-x-struct
  steps:
    - ref:
        kind: Kamelet
        apiVersion: camel.apache.org/v1
        name: insert-field-action
      properties:
        field: fruit
        value: ${header[name]}
    - ref:
        kind: Kamelet
        apiVersion: camel.apache.org/v1
        name: insert-field-action
      properties:
        field: amount
        value: ${header[amount]}
  sink:
    ref:
      kind: KafkaTopic
      apiVersion: kafka.strimzi.io/v1beta2
      name: bookings
    properties:
      brokers: localhost:9092
    dataTypes:
      in:
        format: plain-text
```

oscerd (approved):

PR #1676, opened 2023-10-09
beer-source should not include unwanted HTTP headers

davsclaus:


oscerd (approved):

oscerd:
We could backport to 4.0.x too

davsclaus:
Yes you are welcome to do that, also the chuck norris source can be changed similar

Issue #1698, opened 2023-10-20
Use only camelCase in kamelets yaml


davsclaus:
The old kebab-case for keys in the yaml is deprecated and we should migrate to use camelCase.

For example

    set-header

Should be

    setHeader

And so forth

PR #1724, opened 2023-11-09
Use only camelCase in kamelets yaml - First Batch

oscerd:
Related to #1698


christophd (approved):
YAKS tests should be fixed now on main branch

oscerd:
Rebased, so we could test with YAKS fresh fixed.

PR #2039, opened 2024-05-13
chore: Add Slack Kamelets E2E tests

christophd:
- Add serverUrl property to slack-source.kamelet for better testability reasons
- Use local Slack Http server in tests to verify Slack API calls

### DIFF
--- a/library/camel-kamelets/src/main/resources/kamelets/slack-source.kamelet.yaml
+++ b/library/camel-kamelets/src/main/resources/kamelets/slack-source.kamelet.yaml
@@ -37,6 +37,12 @@ spec:
       - token
     type: object
     properties:
+      serverUrl:
+        title: Server URL
+        description: The Slack API server endpoint URL.
+        type: string
+        default: "https://slack.com"
+        example: "https://slack.com"
       channel:
         title: Channel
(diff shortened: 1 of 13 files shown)


library/camel-kamelets/src/main/resources/kamelets/slack-source.kamelet.yaml
```diff
@@ -37,6 +37,12 @@ spec:
       - token
     type: object
     properties:
+      serverUrl:
```
oscerd, line 40: Do we need to set it as required?

oscerd (commented):
Just a question.

library/camel-kamelets/src/main/resources/kamelets/slack-source.kamelet.yaml
christophd: no it should be optional

christophd: in most cases this will connect to the production Slack API servers (which is also the default in the Camel component). Only tests and maybe a company proxy may use this setting to use a different server URL endpoint

Issue #2261, opened 2024-10-29
Test failure


squakez:
Building is failing with the following error:

```
 Error:  Tests run: 17, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 1.745 s <<< FAILURE! -- in org.apache.camel.kamelets.catalog.KameletsCatalogTest
Error:  org.apache.camel.kamelets.catalog.KameletsCatalogTest.testSupportedHeaders -- Time elapsed: 0.059 s <<< FAILURE!
org.opentest4j.AssertionFailedError: expected: <24> but was: <25>
```

PR #2262, opened 2024-10-29
chore(ci): use internal CRDs spec

squakez:
With this we move the dependency from Camel K CRDs to internal one.


squakez:
Checks are also failing on `main` due to https://github.com/apache/camel-kamelets/issues/2261

PR #2338, opened 2025-01-09
Update kafka-ssl-source.kamelet.yaml

jorgembfigueira:


jorgembfigueira:
@oscerd should i also change the library folder?

oscerd:
> @oscerd should i also change the library folder?

You need to add the title, then you can run mvn clean install from the root folder and it will auto regen the library too. Or you can avoid that and rebuild gh action will open a PR to align it

PR #2369, opened 2025-01-27
chore: Remove obsolete workaround

christophd:
Since Citrus 4.5.2 Camel JBang integrations should be stopped properly on all operating systems so the workaround is not needed anymore


christophd:
Seems like the workaround is still required. Closing and going back to the drawing board

PR #2600, opened 2025-10-02
chore: remove keda descriptors

squakez:
* They were used in an experimental Camel K feature
* Removed as they requires the operator to access to the secrets
* Also, the auth model suggest the usage of a separate credentials from the application's one


squakez:
@oscerd can we merge this? are you aware by any chance of anybody else beside Camel K making use of those annotations?

oscerd:
AFAIK camel-k was the only using these annotations. So that's fine to remove and being neutral.

PR #345, opened 2021-06-09
Space in property "name" of insert-header-action seems to cause java.…

lfabriko:
…lang.IllegalArgumentException

[1] 2021-06-09 19:43:08,803 INFO  [org.apa.cam.imp.eng.AbstractCamelContext] (main) Apache Camel 3.9.0 (camel-1) started in 269ms (build:0ms init:211ms start:58ms)
[1] 2021-06-09 19:43:08,899 INFO  [io.quarkus] (main) camel-k-integration 1.4.0 on JVM (powered by Quarkus 1.13.0.Final) started in 1.604s. Listening on: http://0.0.0.0:8080
[1] 2021-06-09 19:43:08,900 INFO  [io.quarkus] (main) Profile prod activated. 
[1] 2021-06-09 19:43:08,900 INFO  [io.quarkus] (main) Installed features: [camel-attachments, camel-bean, camel-core, camel-k-cloudevents, camel-k-core, camel-k-knative, camel-k-knative-producer, camel-k-loader-yaml, camel-k-runtime, camel-kamelet, camel-platform-http, camel-support-common, camel-timer, camel-yaml-dsl, cdi, mutiny, smallrye-context-propagation, vertx, vertx-web]
I observed following exception on OCP 4.7, community operator Camel K 1.4:

[1] 2021-06-09 19:43:09,858 WARN  [org.apa.cam.imp.eng.DefaultReactiveExecutor] (Camel (camel-1) thread #0 - timer://tick) Error executing reactive work due to a header name cannot contain the following prohibited characters: =,;: \t\r\n\v\f: The Name. This exception is ignored.: java.lang.IllegalArgumentException: a header name cannot contain the following prohibited characters: =,;: \t\r\n\v\f: The Name
[1]     at io.vertx.core.http.impl.HttpUtils.validateHeaderName(HttpUtils.java:766)
[1]     at io.vertx.core.http.impl.HttpUtils.validateHeader(HttpUtils.java:673)
[1]     at io.vertx.core.http.impl.headers.VertxHttpHeaders.add0(VertxHttpHeaders.java:531)

PR #2792, opened 2026-04-13
CAMEL-23300: add connectionString option to couchbase-sink kamelet

Croway:
## Summary
- Expose the new `connectionString` endpoint option (added in [CAMEL-23300](https://issues.apache.org/jira/browse/CAMEL-23300)) as an optional property on the `couchbase-sink` kamelet
- Uses `{{?connectionString}}` so the parameter is only included when explicitly set
- Required for `camel-kafka-connector` daily CI which uses Testcontainers with random port mapping — without `connectionString`, the Couchbase SDK defaults the KV port to 11210 which is unreachable on the host side

## Context
The `camel-couchbase` component added `connectionString` in two commits on `main`:
- `6c2f458` — add `connectionString` endpoint option and switch test-infra to random ports
- `d004589` — bypass Couchbase 8.0 CPU microarchitecture check in test-infra

The camel-kafka-connector daily build has been failing since build #345 because the couchbase-sink kamelet does not expose `connectionString`, so the connector cannot pass it to the endpoint.

## Test plan
- [x] Verified locally: applied this kamelet change + corresponding CKC test changes, ran `CamelSinkCouchbaseITCase` — test passes
- No couchbase integration tests exist in camel-kamelets currently

### DIFF
--- a/kamelets/couchbase-sink.kamelet.yaml
+++ b/kamelets/couchbase-sink.kamelet.yaml
@@ -72,6 +72,10 @@ spec:
         description: The starting id
         type: integer
         default: 1
+      connectionString:
+        title: Connection String
+        description: The full Couchbase SDK connection string (e.g. couchbase://host:port). When set, it takes precedence over hostname extraction for the KV service port.
+        type: string
       autoStartId:
         title: Auto Start Id
         description: Auto Start Id or not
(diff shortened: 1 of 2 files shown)


oscerd (approved):

Croway:
The failures are unrelated

PR #2807, opened 2026-04-29
Fix integration test failures by converting multiline YAML to single-line

cunningt:
Multiline YAML values (value: |) add trailing newlines that break URI encoding and message content processing. Convert all JSON data and input values to single-line quoted strings.

Also update Jira test expectations to match actual JQL ordering (ORDER BY created DESC instead of key desc).

Fixes timeout and URI encoding errors in AWS, transformation, and OpenAPI tests.

### DIFF
--- a/kamelets/databricks-sink.kamelet.yaml
+++ b/kamelets/databricks-sink.kamelet.yaml
@@ -77,7 +77,7 @@ spec:
   - "camel:jackson"
   - "camel:kamelet"
   - "camel:sql"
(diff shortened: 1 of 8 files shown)


oscerd (approved):

oscerd:
There is a failure in the validation

cunningt:
> There is a failure in the validation

Still working through this - I'll put it in draft until I get the CI green

PR #2815, opened 2026-05-01
Remove hardcoded camel / camel-kamelets versions from citrus-application.properties

cunningt:
Starting to work through the integration test failures - there seem to be multiple issues and I think it makes sense to handle them one by one.

The citrus-application.properties file being used contains hardcoded versions of kamelets and camel - we want it to use the versions for testing that are defined in the camel-kamelets pom.xml - so for kamelets, we'd like `${project.version}` and for camel we would like `${camel.version}`.      Make this file a filtered testResource and ensure that the rest of src/test/resources remain unfiltered.

Log the values of citrus.camel.jbang.version and citrus.camel.jbang.kamelets.version to ensure that this fix works in the integration tests.

I expect that after this PR integration tests will still fail as there are other issues (will be handled in separate PRs)

### DIFF
--- a/tests/camel-kamelets-itest/pom.xml
+++ b/tests/camel-kamelets-itest/pom.xml
@@ -186,6 +186,18 @@
     </dependencies>
 
     <build>
+        <testResources>
+            <!-- Standard test resources without filtering to preserve runtime placeholders like ${avro.webhook.server.port} -->
+            <testResource>
+                <directory>src/test/resources</directory>
+                <filtering>false</filtering>
+            </testResource>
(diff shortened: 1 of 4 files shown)


oscerd (approved):

cunningt:
From the CI logs - this is working -

> 2026-05-01 15:28:07.327  INFO 2928 --- [           main] .citrusframework.actions.EchoAction : Camel JBang version: 4.20.0
2026-05-01 15:28:07.327  INFO 2928 --- [           main] .citrusframework.actions.EchoAction : Kamelets version: 4.21.0-SNAPSHOT

PR #2881, opened 2026-06-25
ci: add catalog consistency rules to the Kamelet validator

oscerd:
## Summary

Adds four regression-guard rules to `script/validator`, building on the validator re-arm in #2878 (merged). Each rule encodes a documented catalog convention so drift cannot silently return. With #2878 and #2879 now merged, this PR is a standalone change to `script/validator/validator.go`.

## New rules
1. **Reject deprecated tectonic descriptors** — errors on any `x-descriptors` entry with the `urn:alm:descriptor:com.tectonic.ui` prefix (the catalog standard is `urn:camel:group:credentials`).
2. **Routing convention** — a `source` Kamelet must route to `kamelet:sink`; a `sink` Kamelet must consume from `kamelet:source` (per the README authoring rules). Action Kamelets are unconstrained.
3. **Dependency syntax** — every `spec.dependencies` entry must be `camel:<component>`, `mvn:group:artifact:version`, or `github:apache/...` (enforces the documented licensing/trust constraint).
4. **Catalog-version uniformity** — every Kamelet must share the same `camel.apache.org/catalog.version`, catching a Kamelet added/edited without a version bump.

## Verification
- `go vet` / `gofmt -l`: clean.
- Validator exits `0` across all 250 Kamelets with the rules enabled.
- Each rule was verified to **fire** on an injected violation:
  - `ERROR: property "key" in kamelet "crypto-encrypt-action" uses the deprecated x-descriptor prefix ...`
  - `ERROR: source kamelet "timer-source" must route to "kamelet:sink"`
  - `ERROR: kamelet "timer-source" declares dependency "bogus:timer" that is not a ...`
  - `ERROR: inconsistent camel.apache.org/catalog.version across the catalog: "4.20.0" (1 kamelets), "4.21.0-SNAPSHOT" (249 kamelets)`

## Follow-ups (not in this PR)
- A **"duplicate `x-descriptors`" rule** would be a natural addition — it would have caught the duplicate descriptor @apupier found (fixed separately in #2882).
- Improve `verifyUsedParams` bean / `ref:` detection and replace the hard-coded relative-path exclusion list with an in-Kamelet opt-out annotation.

---
_AI-generated by Claude Code on behalf of Andrea Cosentino (@oscerd)._

### DIFF
--- a/script/validator/validator.go
+++ b/script/validator/validator.go
@@ -72,6 +72,9 @@ func main() {
 	errors = append(errors, verifyDescriptors(kamelets)...)
 	errors = append(errors, verifyDuplicates(kamelets)...)
 	errors = append(errors, verifyUsedParams(kamelets)...)
+	errors = append(errors, verifyTemplateConvention(kamelets)...)
+	errors = append(errors, verifyDependencies(kamelets)...)
+	errors = append(errors, verifyCatalogVersion(kamelets)...)
 
 	for _, err := range errors {
 		fmt.Printf("ERROR: %v\n", err)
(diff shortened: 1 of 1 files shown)


kamelets/crypto-encrypt-action.kamelet.yaml
apupier: is it normal to have 2 times the same value?

oscerd:
@apupier re: your review comment on `crypto-encrypt-action.kamelet.yaml:51` ("is it normal to have 2 times the same value?") — good catch, that was a real bug on my side.

The `key` property already carried `urn:camel:group:credentials`, and my edit was *converting* the deprecated `urn:alm:descriptor:com.tectonic.ui:password` line into `urn:camel:group:credentials` rather than removing it, which produced the duplicate.

Fixed by dropping the deprecated descriptor and keeping the single canonical one. A catalog-wide check turned up the **same** duplicate on `crypto-decrypt-action` and `spring-rabbitmq-sink`/`-source` (identical edit), so all four are fixed — there are now **zero** duplicate `x-descriptors` across the 250 Kamelets. The change lives in #2879 (cherry-picked here); both PRs have been updated and a full `mvn clean install -DskipTests` is green.

_Claude Code on behalf of Andrea Cosentino (@oscerd)_

apupier:
there are conflicts to resolve

oscerd:
Update on the duplicate-descriptor fix: #2879 was merged before the dedup landed, so the correction is now a standalone PR against `main` — **#2882** (removes the redundant `urn:camel:group:credentials` from `crypto-encrypt-action`, `crypto-decrypt-action`, `spring-rabbitmq-sink`/`-source`; a catalog-wide check confirms zero duplicate `x-descriptors`).

This PR (#2881) has been rebased onto `main` and now contains only the validator-rule additions in `script/validator/validator.go`. Thanks again @apupier for catching it.

Issue #929, opened 2022-05-16
A kamelet should define which headers are supported as input and output


davsclaus:
We should start to markup/specify which headers a sink and source kamelet can support as input and output.
For example the ftp sink kamelet should specify that the kamelet outputs `ce-file` as the filename of the FTP file.

The underlying FTP component also provides a set of headers
https://camel.apache.org/components/next/ftp-component.html#_message_headers

From Camel 3.17 then every component now list those in their documentation and metadata json file.

We need to consider whether we want Kamelets to expose these headers too, or we make a rule into kamelets that they output headers that the kamelet has declared.

And we also need to add into the kamelet spec how headers should be declared (ala options today).

davsclaus:
@lburgazzoli @oscerd another requirement for kamelets

oscerd:
We could leverage the Camel catalog from 3.17 ahead, but I guess we'll need to change the Kamelet structure to add some new fields. @lburgazzoli

oscerd:
Only the ones supporting data transformers. Not all

oscerd:
Picking up @cjmamo's 2024 question — *"is this supported now?"* — with a precise answer, since @oscerd's *"only the ones supporting data transformers"* is still the state of things and it is worth writing down exactly what exists.

**Partly supported. A declaration mechanism exists and is modelled end to end; what is missing is that it is coupled to `dataTypes`, nothing enforces it, and the catalog API a consumer would reach for ignores it.**

## What exists

`spec.dataTypes.<in|out>.headers` is a real declaration — a map of header name to `title` / `description` / `type` / `default`. It is modelled in the generated CRDs (`org.apache.camel.v1.kameletspec.DataTypes` carries `Map<String, Headers>`), so it is part of the API, not a convention.

Adoption is **19 of 262 Kamelets**, and all 19 declare it inside a `dataTypes` block: `aws-cloudtrail-source`, `aws-ddb-sink`, `aws-ddb-streams-source`, `aws-s3-source`, `aws-sqs-source`, the Azure storage and Cosmos DB sources, `azure-eventhubs-source`, the Google Calendar / Mail / Sheets / Storage Kamelets, `ibm-cos-source`, `ibm-watson-language-sink` and `slack-source`.

## The three gaps

[...]

PR #2966, opened 2026-08-24
Fix #2952: enforce that the file/ce-file header is a file name in the file-transfer sinks

### LINKED ISSUE #2952: File-transfer sinks: enforce that the file/ce-file header is a file name, not a path
The FTP, FTPS, SFTP and Azure Storage Files sink Kamelets accept an inbound `file` (or `ce-file`) header and use it as the name of the file to write:

- `kamelets/ftp-sink.kamelet.yaml`
- `kamelets/ftps-sink.kamelet.yaml`
- `kamelets/sftp-sink.kamelet.yaml`
- `kamelets/azure-storage-files-sink.kamelet.yaml`

All four share the same copy block:

```yaml
- choice:
    when:
    - simple: "${header[file]}"
      steps:
      - setHeader:
          name: CamelFileName
          simple: "${header[file]}"
    - simple: "${header[ce-file]}"
      steps:
      - setHeader:
          name: CamelFileName
          simple: "${header[ce-file]}"
```

The documented contract for this header is a file *name* — the `ftp-sink`, `ftps-sink` and `azure-storage-files-sink` doc partials say "In the header, you can optionally set the `file` / `ce-file` property to specify the name of the file to upload."

The value is passed through unchanged, so nothing enforces that. A value containing directory separators is resolved by the component relative to the configured `directoryName`, which is wider than what the documentation describes. The templates should reduce the header to a single path segment so the result always lands inside the configured directory.

Two related gaps worth closing in the same change:

- `sftp-sink` has no *Optional Headers* section at all, so its support for this header is undocumented.
- The header is not declared in `spec.definition` for any of the four, so it does not appear in tooling.

Related: #929 (a Kamelet should define which headers are supported as input and output).

oscerd:
Fixes #2952

`ftp-sink`, `ftps-sink`, `sftp-sink` and `azure-storage-files-sink` document the `file` / `ce-file` header as "the name of the file to upload", but passed the value straight through to `CamelFileName`. Nothing enforced that it was a name rather than a path, so a value containing directory separators was resolved relative to the configured `directoryName`.

```diff
             - setHeader:
                 name: CamelFileName
                 simple: "${header[file]}"
+            - setHeader:
+                name: CamelFileName
+                simple: "${file:onlyname}"
```

`${file:onlyname}` is Camel's built-in file language for "the file name with no leading paths" — which is exactly what the docs already promise this header does.

> **Revised after review.** The first version used an inline `replaceAll` chain. @davsclaus rightly objected to a regex embedded in the Kamelet spec and suggested a bean. A bean would have to live in `apache/camel` — `77be42a85` removed the utils module here with *"Remove Camel-Kamelets-utils since it is now in core"* — so it would have blocked this on a Camel release for something the file language already does. The regex is gone either way.

### Verified end to end through a Kamelet template

Not just the expression in isolation — a probe Kamelet mirroring `ftp-sink`'s structure, run on Camel 4.22.0:

| inbound header | file written |
|---|---|
| `file: ../../etc/evil.txt` | `evil.txt`, inside the configured directory |
| `ce-file: sub/dir/report.csv` | `report.csv`, inside the configured directory |
| *(none)* | exchange-id name — default naming convention preserved |

The third row is load-bearing: the normalisation sits **inside** each `choice` branch, because applying it after the choice would set `CamelFileName` to an empty value when no header was supplied, which is not the same as leaving it unset.

### Behaviour change

A header that previously placed a file in a subdirectory — e.g. `reports/2026/data.csv` — now writes `data.csv` into the configured directory instead. If per-message subdirectories are a use case anyone relies on, that deserves an explicit, declared property rather than this header.

One deliberate difference from the earlier revision: `${file:onlyname}` leaves a bare `..` unchanged, where the regex blanked it. That names the parent directory rather than traversing into it, and the file/FTP producers reject it.

### Docs

The header was documented in the `ftp-sink`, `ftps-sink` and `azure-storage-files-sink` partials but **not** in `sftp-sink`'s. All four now state the contract, and `sftp-sink` gains the missing *Optional Headers* section. (The header is still not declared in `spec.definition` for any of the four — that is the broader #929 problem.)

### Verification

- `script/validator` reports no errors
- `mvn clean install` passes from the repository root, full reactor with tests
- Behaviour verified with `camel run` as tabulated above

---

### DIFF
--- a/docs/modules/ROOT/partials/azure-storage-files-sink-description.adoc
+++ b/docs/modules/ROOT/partials/azure-storage-files-sink-description.adoc
@@ -24,3 +24,9 @@ For more information, see the https://learn.microsoft.com/en-us/java/api/overvie
 === Optional Headers
 
 In the headers, you can set the `file` / `ce-file` property to specify the filename to upload. If you do set property in the header, the Kamelet uses the exchange ID as filename.
+
+The value is reduced to a single file name before use: any directory component is
+dropped, so `reports/2026/data.csv` is stored as `data.csv`, and a value made up
+only of dots is discarded in favour of the default naming convention. The file is
+always written inside the configured directory.
+
--- a/docs/modules/ROOT/partials/azure-storage-files-sink-description.adoc
+++ b/docs/modules/ROOT/partials/azure-storage-files-sink-description.adoc
@@ -26,7 +26,6 @@ For more information, see the https://learn.microsoft.com/en-us/java/api/overvie
 In the headers, you can set the `file` / `ce-file` property to specify the filename to upload. If you do set property in the header, the Kamelet uses the exchange ID as filename.
 
 The value is reduced to a single file name before use: any directory component is
-dropped, so `reports/2026/data.csv` is stored as `data.csv`, and a value made up
-only of dots is discarded in favour of the default naming convention. The file is
-always written inside the configured directory.
+dropped, so `reports/2026/data.csv` is stored as `data.csv`. The file is always
+written inside the configured directory.
 
--- a/docs/modules/ROOT/partials/ftp-sink-description.adoc
+++ b/docs/modules/ROOT/partials/ftp-sink-description.adoc
@@ -24,3 +24,8 @@ Requires:
 In the header, you can optionally set the `file` / `ce-file` property to specify the name of the file to upload.
 
 If you do not set the property in the header, the Kamelet uses a default naming convention.
+
+The value is reduced to a single file name before use: any directory component is
+dropped, so `reports/2026/data.csv` is stored as `data.csv`, and a value made up
+only of dots is discarded in favour of the default naming convention. The file is
+always written inside the configured directory.
--- a/docs/modules/ROOT/partials/ftp-sink-description.adoc
+++ b/docs/modules/ROOT/partials/ftp-sink-description.adoc
@@ -26,6 +26,5 @@ In the header, you can optionally set the `file` / `ce-file` property to specify
 If you do not set the property in the header, the Kamelet uses a default naming convention.
 
 The value is reduced to a single file name before use: any directory component is
-dropped, so `reports/2026/data.csv` is stored as `data.csv`, and a value made up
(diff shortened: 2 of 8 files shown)


davsclaus:
Is it not better to have some kind of bean that can do this, and then call that instead of this hacky simple code. This is more safe and does not expose the dirty laundrary in the kamelet spec file

oscerd:
Fair point, and I've changed it — though not to a bean, because Camel already ships something better for this.

`${file:onlyname}` is the built-in file language for "the file name with no leading paths", which is exactly the operation:

```yaml
- setHeader:
    name: CamelFileName
    simple: "${header[file]}"
- setHeader:
    name: CamelFileName
    simple: "${file:onlyname}"
```

No regex, no new class, nothing for us to maintain, and the semantics are owned upstream.

I looked at the bean route first. The precedent is `extract-field-action` (`beans:` + `process: ref:` against `org.apache.camel.component.kafka.transform.ExtractField`), but those helpers no longer live in this repo — `77be42a85` removed the module with *"Remove Camel-Kamelets-utils since it is now in core"*, and `camel-kamelets-utils` stops at 4.8.8 on Maven Central. So a bean would mean a PR to `apache/camel`, a Camel release, and then this one — for something the file language already does.

Verified on 4.22.0, end to end through a Kamelet template rather than just the expression:

| inbound header | file written |
|---|---|
| `file: ../../etc/evil.txt` | `evil.txt`, inside the configured directory |
| `ce-file: sub/dir/report.csv` | `report.csv`, inside the configured directory |
| *(none)* | exchange-id name, i.e. the default convention still applies |

Two things worth your eye:

1. The normalisation has to sit **inside** each `choice` branch. Applying it after the choice sets `CamelFileName` to an empty value when no header was supplied, which is not the same as leaving it unset — that third row is what catches it.
2. `${file:onlyname}` leaves a bare `..` unchanged, where the previous regex blanked it. That names the parent directory rather than traversing into it, and the file/FTP producers reject it, so I took the trade for readability. Say the word if you'd rather it were handled explicitly.

PR #2976, opened 2026-08-24
Fix #2973: strip the remaining destination-affecting HTTP headers in the http sinks

### LINKED ISSUE #2973: Sink Kamelets should strip the dispatch headers they do not deliberately consume
`http-sink` and `http-secured-sink` establish a pattern the rest of the catalog does not follow:

```yaml
- removeHeader:
    name: CamelHttpUri
- setHeader:
    name: CamelHttpMethod
    constant: "{{method}}"
- to: "{{url}}"
```

The Kamelet pins the destination as an operator-bound property and drops the inbound header that would otherwise redirect it. Its doc partial states the intent outright: *"removes any existing CamelHttpUri header to prevent conflicts."*

Across 95 `*-sink` Kamelets, only 7 contain any `removeHeader`/`removeHeaders` step: `http-sink`, `http-secured-sink`, `azure-functions-sink`, `mail-sink`, `solr-sink`, `salesforce-delete-sink`, `salesforce-update-sink`. The rest pass every inbound header through to the component producer.

## The distinction that matters when implementing this

Not every case here is the same kind of problem, and the per-family PRs will get argued past each other if that is not settled up front. The test is **whether the specific header is part of the Kamelet's declared or documented contract** — not whether the family "looks risky".

### Tier 1 — the template consumes nothing, the component honours a header anyway

No `types.*.headers:` declaration, no template mapping, no mention in the doc partial. The template never asked for this input; the component supplies it. These are the genuine gaps.

| family | header | declared headers block | template header mappings |
|---|---|---|---|
| SQL/CQL sinks (9: postgresql, mysql, mariadb, sqlserver, oracle, snowflake, redshift, databricks, cassandra) | `CamelSqlQuery` / `CamelCqlQuery` | 0 | 0 |
| HTTP sinks | `CamelHttpPath`, `CamelHttpQuery` | 0 | 0 |
| broker sinks (jms-*, mqtt*, …) | destination overrides | 0 | 0 |

`CamelSqlQuery` is the clearest: documented by camel-sql as *"takes precedence over the query specified in the endpoint URI"*, while the SQL sink templates map no headers at all and expose the query solely as an operator-bound `{{query}}` property. Nothing about the template invites a wire-supplied query.

### Tier 2 — the header *is* the documented contract

Here a strip would remove a feature, and adding sanitisation is defence in depth behind an operator decision, not a closed hole. Changes to these need to preserve the contract.

- **`aws-ddb-sink`** declares `CamelAwsDdbOperation` in a `types.in.headers:` block, with a description ("Overwrites the operation given in the Json body") and a default of `PutItem`. This is a deliberate, declared interface.
- **`kafka-sink`** maps `key`/`ce-key` and `partition-key`/`ce-partitionkey`, and its doc partial lists them: *"The Kamelet is able to understand the following headers to be set."*
- **`aws-ec2-sink`** maps `instanceIds`/`ce-instanceids` deliberately.
- **`exec-sink`** maps `args`/`ce-args`; the security model names this behaviour explicitly as by design.

> **Correction to the first version of this issue:** I cited aws-ddb-sink's "Overwrites the operation" text as evidence of the problem. It is the opposite — that text is a declared header contract in `spec.types.in.headers`. aws-ddb-sink belongs in tier 2, not tier 1.

### Tier 3 — mixed inside a single Kamelet

The common and easiest-to-get-wrong case: a template deliberately consumes one header from a family and silently passes the rest.

- **`exec-sink`** — `args`/`ce-args` is tier 2; `CamelExecCommandExecutable`, `CamelExecCommandWorkingDir` and `CamelExecCommandOutFile` are tier 1. The Kamelet is named for running a command, but not for letting the sender choose *which binary*.

  **However — the executable override does not reproduce.** The component catalog documents `CamelExecCommandExecutable` as *"The name of the system command that will be executed. Overrides executable in the URI."* On Camel 4.21 it is ignored, tested both ways:

  | endpoint | header | command that actually ran |
  |---|---|---|
  | `exec:hostname` | `whoami` | `hostname` |
  | `exec:whoami` | `hostname` | `whoami` |

  The header also survived the call unconsumed, and `camel-exec` reads it with `getAndRemoveHeader`, so it was never looked at. Either this is a `camel-exec` documentation/behaviour mismatch worth raising upstream, or it differs between 4.21 and 4.22 — I tested with the 4.21 JBang CLI against the 4.22 catalog metadata, so someone should confirm on 4.22 before treating the exec strip as load-bearing. Stripping the headers is harmless either way, but this is not the open door the original report described.
- **`kafka-sink`** — `key` and `partition-key` are tier 2; the topic is not. The override header is `CamelKafkaOverrideTopic` (`KafkaConstants#OVERRIDE_TOPIC`), documented as *"The topic to which send the message (override and takes precedence)"*.
- **`aws-ec2-sink`** — `instanceIds` is tier 2; the operation header is tier 1 (unlike `aws-ddb-sink`, it declares no headers block).

## Reachability

Whether a given inbound header reaches the exchange depends on the source in front of the sink — several Camel source components apply a header filter on ingest that drops `Camel`-prefixed names.

**Correction to the first version of this issue.** It claimed `kafka.OVERRIDE_TOPIC` was not `Camel`-prefixed and therefore reached the sink regardless of ingest filtering. That is the pre-4.x name. In Camel 4.22 it is `CamelKafkaOverrideTopic`, and #2871 records that the rename is exactly what closed the HTTP leak: *"With the migration to `CamelKafka*` names, the HTTP component now correctly filters them."*

With that corrected, **every** dispatch-override header in these families is `Camel`-prefixed, so all of them are subject to source-side ingest filtering. (The other candidate for a non-prefixed name, a bare `operation` on the Elasticsearch/OpenSearch sinks, does not exist either — the header is `CamelElasticsearchOperation`.)

That makes the whole of this issue defence in depth rather than a set of open holes: it closes the "if the header arrives, it redirects the sink" half, while the "can it arrive" half is already handled at the source for the common HTTP and Kafka topologies. The residual path is a peer that can set properties directly on the operator's broker, which is a much narrower position than an arbitrary upstream sender.

Worth doing, worth doing accurately, and **not** worth describing as a vulnerability in the PRs that implement it.

## Proposal

Extend the `http-sink` discipline per family: each sink strips the dispatch headers of its own component that the template does not deliberately consume, immediately before `to:`. Tier 1 first — those are strips with no contract to preserve.

Given the reachability correction above, the broker slice is the lowest-value part of this and should be prioritised accordingly — after tier 3, not before it. Tier 3 needs the deliberate header kept and the rest dropped. Tier 2 needs no change beyond documenting what is already intended.

Only 17 Kamelets currently declare a `types.*.headers:` block, and `aws-ddb-sink` is the only sink among them. Declaring the intended header interface as those 17 do would make this distinction mechanical rather than a judgement call each time — that is #929, and it is worth doing first if anyone is picking up both.

Related: #2871 (Kafka kamelets should use kamelet properties instead of leaking Camel internal headers) covers the Kafka slice. Also #929.

oscerd:
Partially addresses #2973 — the tier-1 HTTP slice. SQL/CQL is in a sibling PR; broker sinks and the tier-3 mixed cases (`exec-sink`, `kafka-sink`, `aws-ec2-sink`) are still open.

`http-sink` and `http-secured-sink` already removed `CamelHttpUri` before dispatch, and the doc partial states the intent: *"removes any existing CamelHttpUri header to prevent conflicts."* But `camel-http` builds the request URI from several more headers, none of which were stripped.

```diff
       - removeHeader:
           name: CamelHttpUri
+      - removeHeader:
+          name: CamelHttpPath
+      - removeHeader:
+          name: CamelHttpQuery
+      - removeHeader:
+          name: CamelHttpRawQuery
+      - removeHeader:
+          name: CamelRestHttpUri
+      - removeHeader:
+          name: CamelRestHttpQuery
       - setHeader:
           name: CamelHttpMethod
```

### Verified against a local listener

The endpoint was configured for `/intended`, and the route set `CamelHttpPath: /hijacked` plus `CamelHttpQuery: injected=1`:

| template | path the server received |
|---|---|
| current `http-sink` | `/intended/hijacked?injected=1` |
| this PR | `/intended` |

Worth noting these headers **append to** the operator's path rather than replacing it — `CamelHttpPath` is documented as *"used to build the request URI with the HTTP_URI"*. I would have described this wrongly from the documentation alone, which is why it was worth running.

### Why named headers rather than a `CamelHttp*` pattern

The obvious `removeHeaders: pattern: "CamelHttp*"` is wrong in both directions:

- it would also drop `CamelHttpCharacterEncoding` and `CamelHttpProtocolVersion`, which are legitimate things for a route to set;
- it would still miss `CamelRestHttpUri` and `CamelRestHttpQuery`, which do not start with `CamelHttp`.

The named list covers exactly the headers `camel-http` uses to build the request URI. `Content-Type` and `Content-Encoding` are untouched.

### Tier

Tier 1 in #2973's taxonomy: these templates declare no `types.*.headers:` block and map no headers of their own, so nothing is being taken away. The Kamelet's whole point is that `{{url}}` is operator-bound.

### Verification

- `script/validator` reports no errors
- `mvn verify` passes
- Behaviour verified with `camel run` against a local HTTP listener as tabulated

---

### Framing

Please read this as defence in depth, not as closing an open hole. Every dispatch-override header in these families is `Camel`-prefixed and therefore subject to source-side ingest filtering — see the corrected *Reachability* section of #2973. This PR closes the "if the header arrives, it redirects the sink" half; the "can it arrive" half is already handled at the source for the common topologies.

### DIFF
--- a/kamelets/http-secured-sink.kamelet.yaml
+++ b/kamelets/http-secured-sink.kamelet.yaml
@@ -105,6 +105,16 @@ spec:
       steps:
       - removeHeader:
           name: CamelHttpUri
+      - removeHeader:
+          name: CamelHttpPath
+      - removeHeader:
+          name: CamelHttpQuery
+      - removeHeader:
+          name: CamelHttpRawQuery
+      - removeHeader:
+          name: CamelRestHttpUri
+      - removeHeader:
+          name: CamelRestHttpQuery
       - setHeader:
           name: CamelHttpMethod
           constant: "{{method}}"
--- a/kamelets/http-sink.kamelet.yaml
+++ b/kamelets/http-sink.kamelet.yaml
@@ -57,6 +57,16 @@ spec:
       steps:
       - removeHeader:
           name: CamelHttpUri
+      - removeHeader:
+          name: CamelHttpPath
+      - removeHeader:
+          name: CamelHttpQuery
+      - removeHeader:
+          name: CamelHttpRawQuery
+      - removeHeader:
+          name: CamelRestHttpUri
+      - removeHeader:
+          name: CamelRestHttpQuery
       - setHeader:
           name: CamelHttpMethod
           constant: "{{method}}"


davsclaus:
You could also use removeHeaders and use CamelHttp* as the pattern

davsclaus (approved):

PR #2978, opened 2026-08-24
Fix #2973: strip non-contract dispatch headers in aws-ec2-sink

### LINKED ISSUE #2973: Sink Kamelets should strip the dispatch headers they do not deliberately consume

oscerd:
Partially addresses #2973 — the tier-3 slice (templates that deliberately consume *one* header from their component's family and pass the rest through). HTTP is #2976, SQL/CQL is #2977. Broker sinks remain, and per the corrected reachability note in the issue they are the lowest-value part.

After review this PR is down to a **single file**.

| Kamelet | kept (documented interface) | stripped |
|---|---|---|
| `aws-ec2-sink` | `instanceIds` / `ce-instanceids` → `CamelAwsEC2InstancesIds`, plus the `NextToken` / `MaxResults` pagination pair | rest of `CamelAwsEC2*` |

`aws-ec2-sink` is the clearest shape in the slice: it pins `operation: describeInstancesStatus` — a read-only call — while deliberately letting the sender name instances. A `CamelAwsEC2Operation` header could turn that pinned read into any other EC2 operation the component supports.

Following review, the strip is a single `removeHeaders` **pattern** placed *before* the Kamelet sets its own headers, rather than an enumerated `removeHeader` list after:

```yaml
      - removeHeaders:
          pattern: "CamelAwsEC2*"
          excludePattern: "CamelAwsEC2(NextToken|MaxResults)"
      - choice:            # instanceIds / ce-instanceids -> CamelAwsEC2InstancesIds
```

Stripping up-front means the Kamelet's own contract is re-applied afterwards and needs no exception list, and future `CamelAwsEC2*` keys are covered without another PR.

The single `excludePattern`: `CamelAwsEC2NextToken` and `CamelAwsEC2MaxResults` are the pagination headers that the pinned `describeInstancesStatus` operation actually reads. They are functional rather than dispatch-controlling, so stripping them would be a silent regression with no security benefit.

The strip carries a comment saying what the contract is, so the distinction survives future edits.

### `kafka-sink` — the change was wrong

An earlier revision stripped `CamelKafkaOverrideTopic` / `CamelKafkaOverrideTimestamp` in `kafka-sink`. **That was wrong and has been reverted.**

Three shipped Kamelets exist solely to set `CamelKafkaOverrideTopic` for a downstream `kafka-sink`: `regex-router-action`, `timestamp-router-action` and `message-timestamp-router-action` (wrapping `RegexRouter`, `TimestampRouter` and `MessageTimestampRouter` respectively). `tests/.../kafka/kafka-router-pipe.yaml` is a Pipe of exactly that shape, whose Citrus assertion consumes from the **overridden** topic — the sink's bound `topic: dummy` is deliberately a throwaway.

Reproduced against the removed revision:

```
BEFORE SINK -> CamelKafkaOverrideTopic=prefix-my-topic
PROBE       : Exchange[Headers: {CamelKafkaTopic=my-topic}, BodyType: String]
```

Header-driven topic override is a designed catalog feature, not an unintended pass-through, and the strip silently disabled three Kamelets. This also rules out a blanket `CamelKafka*` pattern there: it would additionally strip `CamelKafkaKey` (set by `value-to-key-action`) and the consumer-side `CamelKafkaManualCommit` / `Offset` / `LastRecordBeforeCommit` that a `kafka-source` → `kafka-sink` Pipe carries.

### `exec-sink` — the change is a no-op today

The original report treated `CamelExecCommandExecutable` as an open door. **I could not reproduce that.** Re-tested against a real Camel **4.22.0** runtime, `camel-exec` honours none of its documented `CamelExecCommand*` headers — not the executable, and not `CamelExecCommandArgs` either, with an identical route shape as a control:

| how the argument is passed | stdout | header afterwards |
|---|---|---|
| `exec:echo?args=URIARGS-WORK` (URI parameter) | `URIARGS-WORK` | n/a |
| `CamelExecCommandArgs` header, `exec:echo` | *(empty)* | still present, unconsumed |

`exec-sink`'s own documented `args` / `ce-args` interface is dead code as a result. Filed separately as #2980.

That made the `exec-sink` strip purely forward-looking — it would only start doing anything if `camel-exec` were fixed upstream. Rather than carry a no-op in the catalog until then, it has been dropped from this PR. It can be revisited alongside #2980.

## Scope note

`CamelAwsEC2Operation` is **not** verified end-to-end — that needs a live AWS account. It is documented as overriding, and unlike the exec case there is no evidence against that, but treat it as documentation-based rather than demonstrated. The HTTP (#2976) and SQL (#2977) slices are the two proved end-to-end.

## Framing

Defence in depth, not an open hole. Every header here is `Camel`-prefixed and so subject to source-side ingest filtering — see the corrected *Reachability* section of #2973.

## Verification

- `script/validator` reports no errors; `script/generator` produces no doc changes (template steps are not rendered into docs)
- `mvn clean install` passes from the repository root
- Header behaviour probed with `camel run` against the real template, terminal `to:` swapped for a `log:` so the headers at dispatch time are visible:

```
PROBEEC2 : Exchange[Headers: {CamelAwsEC2InstancesIds=i-123, CamelAwsEC2MaxResults=10,
                              CamelAwsEC2NextToken=PAGE2, instanceIds=i-123, keepMe=yes}]
```

`CamelAwsEC2Operation` / `ImageId` / `InstanceType` are gone; the Kamelet's own contract header survives; the pagination pair and unrelated headers are untouched.

## CI gap found along the way

`kafka-router-pipe-test` **passed** on an earlier head despite provably breaking on it, so the Citrus itests are not running against the catalog built in the same run. That is a gate problem, not a change in this PR — filed separately as #2984.

---

### DIFF
--- a/kamelets/aws-ec2-sink.kamelet.yaml
+++ b/kamelets/aws-ec2-sink.kamelet.yaml
@@ -110,6 +110,10 @@ spec:
             - setHeader:
                 name: CamelAwsEC2InstancesIds
                 simple: "${header[ce-instanceids]}"
+      # instanceIds / ce-instanceids above are this Kamelet's documented interface
+      # and are kept. The operation is pinned to describeInstancesStatus below.
+      - removeHeader:
+          name: CamelAwsEC2Operation
       - to:
           uri: "aws2-ec2:ec2-route"
           parameters:
--- a/kamelets/aws-ec2-sink.kamelet.yaml
+++ b/kamelets/aws-ec2-sink.kamelet.yaml
@@ -98,6 +98,15 @@ spec:
     from:
       uri: kamelet:source
       steps:
+      # Strip the whole camel-aws2-ec2 header family up-front, before this
+      # Kamelet applies its own contract below: the operation is pinned to the
+      # read-only describeInstancesStatus and instanceIds / ce-instanceids is
+      # the only interface the definition advertises. Using a pattern also
+      # covers CamelAwsEC2* keys the component may add in the future. The
+      # pagination headers that describeInstancesStatus itself reads are kept.
+      - removeHeaders:
+          pattern: "CamelAwsEC2*"
+          excludePattern: "CamelAwsEC2(NextToken|MaxResults)"
       - choice:
           when:
           - simple: "${header[instanceIds]}"
@@ -110,10 +119,6 @@ spec:
             - setHeader:
                 name: CamelAwsEC2InstancesIds
                 simple: "${header[ce-instanceids]}"
-      # instanceIds / ce-instanceids above are this Kamelet's documented interface
-      # and are kept. The operation is pinned to describeInstancesStatus below.
-      - removeHeader:
-          name: CamelAwsEC2Operation
       - to:
           uri: "aws2-ec2:ec2-route"
           parameters:
--- a/kamelets/exec-sink.kamelet.yaml
+++ b/kamelets/exec-sink.kamelet.yaml
@@ -47,6 +47,13 @@ spec:
     from:
       uri: kamelet:source
       steps:
+      # Strip the whole camel-exec header family up-front, before this Kamelet
+      # applies its own contract below: the executable is pinned to
+      # exec:{{executable}} and args / ce-args is the only interface the
+      # definition advertises. Using a pattern also covers CamelExec* keys the
+      # component may add in the future.
+      - removeHeaders:
+          pattern: "CamelExec*"
       - choice:
           when:
           - simple: "${header[args]}"
@@ -59,14 +66,6 @@ spec:
             - setHeader:
                 name: CamelExecCommandArgs
                 simple: "${header[ce-args]}"
-      # args / ce-args above are this Kamelet's documented interface and are kept.
-      # These three would override the operator-pinned exec:{{executable}}.
-      - removeHeader:
-          name: CamelExecCommandExecutable
-      - removeHeader:
-          name: CamelExecCommandWorkingDir
-      - removeHeader:
-          name: CamelExecCommandOutFile
       - to:
           uri: "exec:{{executable}}"
       - setBody:
--- a/kamelets/exec-sink.kamelet.yaml
+++ b/kamelets/exec-sink.kamelet.yaml
@@ -59,6 +59,14 @@ spec:
             - setHeader:
                 name: CamelExecCommandArgs
                 simple: "${header[ce-args]}"
+      # args / ce-args above are this Kamelet's documented interface and are kept.
+      # These three would override the operator-pinned exec:{{executable}}.
+      - removeHeader:
+          name: CamelExecCommandExecutable
+      - removeHeader:
+          name: CamelExecCommandWorkingDir
+      - removeHeader:
+          name: CamelExecCommandOutFile
       - to:
           uri: "exec:{{executable}}"
       - setBody:
--- a/kamelets/exec-sink.kamelet.yaml
+++ b/kamelets/exec-sink.kamelet.yaml
@@ -47,13 +47,6 @@ spec:
     from:
       uri: kamelet:source
       steps:
-      # Strip the whole camel-exec header family up-front, before this Kamelet
-      # applies its own contract below: the executable is pinned to
-      # exec:{{executable}} and args / ce-args is the only interface the
-      # definition advertises. Using a pattern also covers CamelExec* keys the
-      # component may add in the future.
-      - removeHeaders:
-          pattern: "CamelExec*"
       - choice:
           when:
           - simple: "${header[args]}"
--- a/kamelets/kafka-sink.kamelet.yaml
+++ b/kamelets/kafka-sink.kamelet.yaml
@@ -143,6 +143,12 @@ spec:
             - setHeader:
                 name: CamelKafkaPartitionKey
                 simple: "${header[ce-partitionkey]}"
+      # key / partition-key above are this Kamelet's documented interface and are
+      # kept. The topic is pinned to {{topic}} below.
+      - removeHeader:
+          name: CamelKafkaOverrideTopic
+      - removeHeader:
+          name: CamelKafkaOverrideTimestamp
       - to:
           uri: "kafka:{{topic}}"
           parameters:
--- a/kamelets/kafka-sink.kamelet.yaml
+++ b/kamelets/kafka-sink.kamelet.yaml
@@ -143,12 +143,6 @@ spec:
             - setHeader:
                 name: CamelKafkaPartitionKey
                 simple: "${header[ce-partitionkey]}"
-      # key / partition-key above are this Kamelet's documented interface and are
-      # kept. The topic is pinned to {{topic}} below.
-      - removeHeader:
-          name: CamelKafkaOverrideTopic
-      - removeHeader:
-          name: CamelKafkaOverrideTimestamp
       - to:
           uri: "kafka:{{topic}}"
           parameters:


davsclaus (commented):
if the kamelets should not allow any existing headers, you could just remove with pattern to ensure all are removed before setting new headers. Then its safe for new introduced header keys in the future

davsclaus (approved):

oscerd:
@davsclaus thanks — adopted, and chasing it down changed what this PR does, so flagging that up front.

## Adopted: `exec-sink` and `aws-ec2-sink`

Both now use a single pattern strip placed **before** the Kamelet sets its own headers, as you suggested:

```yaml
      - removeHeaders:
          pattern: "CamelExec*"
      - choice:            # args / ce-args -> CamelExecCommandArgs
```

```yaml
      - removeHeaders:
          pattern: "CamelAwsEC2*"
          excludePattern: "CamelAwsEC2(NextToken|MaxResults)"
      - choice:            # instanceIds / ce-instanceids -> CamelAwsEC2InstancesIds
```

Moving the strip ahead of the `choice` removes the need to enumerate exceptions for the headers the Kamelet itself sets, and it caught keys the enumerated version had missed — `CamelExecCommandTimeout`, `CamelExecExitValues`, `CamelExecCommandLogLevel`.

The one `excludePattern`: `aws-ec2-sink` pins `operation: describeInstancesStatus`, and `CamelAwsEC2NextToken` / `CamelAwsEC2MaxResults` are the pagination headers that operation actually reads. They are functional rather than dispatch-controlling, so stripping them would have been a silent regression with no security benefit.

Verified with Camel JBang against the real templates (terminal `to:` swapped for a `log:` so the headers at dispatch time are visible):

```
PROBEEXEC : Exchange[Headers: {args=hello-args, CamelExecCommandArgs=hello-args, keepMe=yes}]
PROBEEC2  : Exchange[Headers: {CamelAwsEC2InstancesIds=i-123, CamelAwsEC2MaxResults=10,
                               CamelAwsEC2NextToken=PAGE2, instanceIds=i-123, keepMe=yes}]
```

`CamelExecCommandExecutable` / `WorkingDir` / `OutFile` / `Timeout` and `CamelAwsEC2Operation` / `ImageId` / `InstanceType` are gone; each Kamelet’s own contract header survives; unrelated headers are untouched.

## Not adopted: `kafka-sink` is reverted

Your question — *should this Kamelet allow any existing headers?* — is the one that broke this half. For `kafka-sink` the answer turns out to be **yes**, and the original enumerated strip was already wrong.

Three shipped Kamelets exist for the sole purpose of setting `CamelKafkaOverrideTopic` for a downstream `kafka-sink`:

| Kamelet | bean |
|---|---|
| `regex-router-action` | `kafka.transform.RegexRouter` |
| `timestamp-router-action` | `kafka.transform.TimestampRouter` |
| `message-timestamp-router-action` | `kafka.transform.MessageTimestampRouter` |

There is a Pipe test for exactly that composition — `tests/.../kafka/kafka-router-pipe.yaml` is `webhook-source -> timestamp-router-action -> ... -> kafka-sink (topic: dummy)`, and the Citrus assertion consumes from the **overridden** topic, with `dummy` deliberately a throwaway.

Reproduced locally against the PR’s own template:

```
BEFORE SINK -> CamelKafkaOverrideTopic=prefix-my-topic
PROBE       : Exchange[Headers: {CamelKafkaTopic=my-topic}, BodyType: String]
```

So the premise in that file’s comment — *“the topic is pinned to `{{topic}}`”* — was simply false. Header-driven topic override is a designed catalog feature, and the strip silently disabled three Kamelets. `kafka-sink` is now byte-identical to `main`.

This also rules out a blanket `CamelKafka*` there: it would additionally strip `CamelKafkaKey` (set by `value-to-key-action`) and the consumer-side `CamelKafkaManualCommit` / `Offset` / `LastRecordBeforeCommit` that a `kafka-source` -> `kafka-sink` Pipe carries.

## One thing worth a maintainer’s eye

`kafka-router-pipe-test` **passed** on the previous head despite the above, so the itests are not running against the catalog built in the same run. Filed separately as #2984 — no change here.

Since the changeset moved (`kafka-sink` dropped, `exec-sink` / `aws-ec2-sink` reshaped), happy to have another look when you have a moment.

---

oscerd:
Narrowed further: `exec-sink` is dropped too, so this PR is now a **single file** — `aws-ec2-sink`.

Reasoning is the sceptical note from the description: `camel-exec` honours none of its documented `CamelExecCommand*` headers on 4.22.0 (#2980), so stripping them was a no-op that would only start doing anything if that is fixed upstream. Not worth carrying in the catalog until then — it can be revisited alongside #2980.

What remains stands on its own and takes effect today: `aws-ec2-sink` pins `operation: describeInstancesStatus`, a read-only call, and a `CamelAwsEC2Operation` header could turn that into any other EC2 operation the component supports.

```yaml
      - removeHeaders:
          pattern: "CamelAwsEC2*"
          excludePattern: "CamelAwsEC2(NextToken|MaxResults)"
      - choice:            # instanceIds / ce-instanceids -> CamelAwsEC2InstancesIds
```

@davsclaus your pattern suggestion is what this file uses — strip up-front, let the Kamelet re-apply its own contract afterwards, no exception list needed and future `CamelAwsEC2*` keys covered. The one `excludePattern` keeps `NextToken` / `MaxResults`, the pagination headers the pinned operation actually reads.

Title and description updated. Since the scope has moved twice since your approval, a fresh look would be welcome before this goes in.

---

davsclaus (approved):

PR #3008, opened 2026-09-07
Fix #542: add a Citrus integration test for couchbase-source

### LINKED ISSUE #542: Create Couchbase Source Kamelet

oscerd:
Follow-up to #3007, which shipped `couchbase-source` without a test on the grounds that the project's Citrus toolchain had no Couchbase container. That was true only in the sense that there is no *built-in* one — the generic container works fine, it just needs setting up by hand.

## The test

```
testcontainers: start: container (couchbase/server:community-7.6.2)
  -> initCluster.groovy   (cluster init, bucket, primary index, seed one document)
  -> camel: cli: run      (couchbase-source -> log-sink)
  -> camel: cli: verify   (logMessage: "hello-from-couchbase")
```

Same shape as the existing `mongodb-source-route` test, so it should read familiarly.

## Three things that were not optional

Recording these because each cost a full run to find, and the next person adding a container-based test will hit at least one.

**Ports are bound one-to-one.** The Couchbase SDK bootstraps from the management port and then reconnects using the addresses the *cluster* advertises. With a remapped host port the client bootstraps and then cannot reach the data service. Hence explicit `8091:8091`, `8093:8093`, `11210:11210` rather than `exposedPorts`.

**The wait strategy needs a full-line regex.** Testcontainers matches log wait patterns against the entire line, so `"Starting Couchbase Server"` never matches the actual line and the container times out after five minutes. It has to be `".*Starting Couchbase Server.*"`.

**No Groovy GStrings in the script.** Citrus resolves dollar-brace expressions in these scripts as its own test variables, so a GString fails with `Unknown variable 'user'`. The script uses string concatenation throughout — including, eventually, in the comment that explains why.

## Why the statement looks the way it does

```properties
couchbase.statement=SELECT META().id AS __id, * FROM _default
```

Not the obvious `SELECT * FROM <bucket>`, because both halves matter:

- `CouchbaseConsumer` reads `row.getString("__id")` and **skips any row without it**, logging a warning rather than failing. A query without `META().id AS __id` yields a source that polls happily and emits nothing.
- The consumer runs the query through `scope.query(...)`, so the keyspace is resolved as `<bucket>._default.<what-you-wrote>`. `FROM <bucket>` becomes `<bucket>._default.<bucket>` and fails with `Keyspace not found`.

I found both by running the Kamelet against a live cluster, and they mean **the example currently shipped in `couchbase-source` is wrong** — `SELECT * FROM \`travel-sample\` LIMIT 10` would silently return nothing. That is my error from #3007. I am fixing it in a separate PR rather than folding a Kamelet change into a test PR.

## Verification

Passes locally, twice through (test and verify phases), about 35 seconds each:

```
Tests run: 1, Failures: 0, Errors: 0, Skipped: 0 -- in CouchbaseIT
  ✔ SUCCESS (38088ms) couchbase-source-route-test
```

`mvn clean install` passes with tests from the repository root.

## For reviewers

**The image is large** — `couchbase/server:community-7.6.2` is about 1.7 GB, materially bigger than redpanda or floci. On a cold CI cache that is a real cost for one test, and worth weighing against the value of covering this Kamelet.

**I did not add `camel.apache.org/kamelet.verified: "true"`.** The contributor guide ties that label to having passing behaviour tests, but it is applied to only 21 Kamelets and notably not to `mongodb-source`, which does have a test — so it looks like a curation decision rather than an automatic consequence. Happy to add it if you want it applied here.

---

### DIFF
--- a/tests/camel-kamelets-itest/src/test/java/CouchbaseIT.java
+++ b/tests/camel-kamelets-itest/src/test/java/CouchbaseIT.java
@@ -1,0 +1,33 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import java.util.stream.Stream;
+
+import org.citrusframework.api.common.TestLoader;
+import org.citrusframework.junit.jupiter.CitrusSupport;
+import org.citrusframework.junit.jupiter.CitrusTestFactory;
+import org.citrusframework.junit.jupiter.CitrusTestFactorySupport;
+import org.junit.jupiter.api.DynamicTest;
+
+@CitrusSupport
+public class CouchbaseIT {
+
+    @CitrusTestFactory
+    public Stream<DynamicTest> couchbase() {
+        return CitrusTestFactorySupport.factory(TestLoader.YAML).packageScan("couchbase");
+    }
+}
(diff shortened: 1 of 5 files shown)


oscerd:
CI is green, and the new test genuinely ran rather than being skipped:

```
Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 61.15 s -- in CouchbaseIT
```

Adding the cost figure to go with the image-size caveat above, since it is the number that matters for the decision: **61 seconds on the CI runner**, against 35 locally — the difference being the image pull on a cold cache.

For context, that puts it mid-pack rather than at the top:

```
CommonIT     398.5 s
AwsIT        408.3 s
AvroIT        86.99 s
KafkaIT       80.15 s
ProtobufIT    79.20 s
JiraIT        69.43 s
CouchbaseIT   61.15 s
MongoIT       50.78 s
```

So the 1.7 GB image costs about a minute of a suite that already runs for well over twenty. Still your call whether one Kamelet is worth that, but it is a smaller marginal cost than the image size alone suggests.

---

Issue #1173, opened 2022-11-22
Kamelet validator prevents @BeanInject usage


christophd:
I tried to use `@BeanInject` in camel-kamelets-utils in order to autowire beans that are defined in Kamelet template section. The Kamelet validation scripts prevents this with `parameter %q is declared in kamelet %q but never used`

This is because the bean is declared in Kamelet but not used with `#bean:{{myBean}}` syntax due to @BeanInject.

Is @BeanInject something we want to support? We would need to change the Kamelet validation tool accordingly

oscerd:
I don't have strong opininion on this. I would prefer to keep the stuff simple, but relaxing validation for having some more flexibility in relation to beans is fine.

oscerd:
The premise of this issue no longer exists, so I think it can be closed — but the underlying tension it identified is still there, in a different form, and is worth recording before it goes.

## `camel-kamelets-utils` is gone

The `@BeanInject` case was about autowiring beans from `camel-kamelets-utils`, which this repository no longer ships. `77be42a85` ("Remove Camel-Kamelets-utils since it is now in core", 2024-10-11) moved it into Camel core.

Verified against current `main`:

- no Java sources outside `tests/` and `library/`
- no reference to `camel-kamelets-utils` anywhere in the tree
- no occurrence of `BeanInject` at all

Kamelets are YAML templates only now, so there is no place left in this repository for the `@BeanInject` usage this issue was about. Nothing to support.

## What the validator actually does today

The rule is in `verifyUsedParams` (`script/validator/validator.go`), and it is symmetric — it errors both on a parameter used but not declared, and on one declared but never used:

```go
errors = append(errors, fmt.Errorf("parameter %q is declared in kamelet %q but never used", p, k.Kamelet.Name))
```

The problem you hit — a parameter that is legitimately consumed by something other than direct interpolation into the template — is real and did not go away with the utils module. It is currently handled by a **hardcoded list of 13 exempt files** at the top of that function:

```go
if k.FileName != "../../kamelets/azure-storage-blob-source.kamelet.yaml" &&
   k.FileName != "../../kamelets/aws-s3-event-based-source.kamelet.yaml" &&
   ... eleven more ...
```

So the answer to "would we need to change the Kamelet validation tool" turned out to be yes, and the change that happened was an exclusion list rather than an understanding of indirect usage.

[...]

PR #3026, opened 2026-09-14
Fix #2303: add a Ceph event based source Kamelet

### LINKED ISSUE #2303: Ceph: support bucket notifications?
The ceph source and sink from kamelet seems to be dealing with objects and not notifications.

We have something like this currently in Knative: https://github.com/knative-extensions/eventing-ceph/blob/main/pkg/adapter/adapter.go#L91

However, with the general move towards `eventing-integrations` and using Kamelets inside, I was wondering if it is an option to add bucket notifications to the kamelet catalog.

this can do much more than what the current eventing-ceph code does. e.g. configuring the notifications, and maybe integrate with the ceph-source so that objects will be consumed based on the notifications.
creating a notification based mechanism instead of polling (as the current ceph-source does) is much more efficient.

oscerd:
Fixes #2303.

Builds what @yuvalif asked for and confirmed on the issue. Ceph RGW **pushes** bucket notifications rather than being polled, and his deployment uses Kafka for that — so this consumes those notifications and optionally fetches the object each one refers to, mirroring what `aws-s3-event-based-source` does for SQS.

## The parsing could not be shared with the AWS Kamelet

This is the part that justifies a separate Kamelet rather than a parameter on the existing one. Ceph emits the **classic S3 envelope** — a `Records` array — while `aws-s3-event-based-source` parses the **EventBridge** shape:

| | event name | bucket | key |
|---|---|---|---|
| **Ceph** | `Records[0].eventName` | `Records[0].s3.bucket.name` | `Records[0].s3.object.key` |
| **AWS** | `$.detail.reason` | `$.detail.bucket.name` | `$.detail.object.key` |

Completely different paths. There is no shared parsing to factor out.

## What it reuses

Object fetching goes through the **existing Ceph path** — `aws2-s3` with `uriEndpointOverride` against the RGW endpoint, exactly as `ceph-source` and `ceph-sink` already do. No new component is involved. The Kafka properties mirror `kafka-source` (`saslAuthType` / `saslUsername` / `saslPassword`, `autoOffsetReset`, `consumerGroup`) so the two read alike.

## What stays out of scope, and why

Configuring the notification is **not** something this Kamelet does, and the description says so. It needs two API calls Camel cannot currently make:

- `PutBucketNotificationConfiguration` — absent from `camel-aws2-s3`'s 27 operations
- an SNS-style `CreateTopic` carrying `push-endpoint: kafka://...` — `camel-aws2-sns` can point at RGW via `uriEndpointOverride` but exposes no operation enum and no topic attributes, so the push endpoint cannot be expressed

That keeps this consistent with the position in the thread that the component will not grow Ceph-specific configuration. The Kamelet consumes a notification someone else configured.

## Verification

Ran against a live broker with the event payload taken verbatim from the [Ceph documentation](https://docs.ceph.com/en/latest/radosgw/notifications/#events) that @yuvalif linked.

**Consuming** — the notification arrives and is emitted unchanged:

```
CEPHEVENT >>> {"Records":[{"eventVersion":"2.1","eventSource":"ceph:s3", ... "eventName":"ObjectCreated:Put" ...}]}
```

**Parsing**, with `getObject` enabled and the fetch swapped for a log so the extracted values are visible:

```
PARSED event=[ObjectCreated:Put] bucket=[mybucket1] key=[myimage1.jpg]
```

All three paths resolve against the real envelope.

`script/validator` reports no errors, `script/generator` adds the `nav.adoc` entry, `mvn clean install` passes with tests from the repository root.

## One change outside the Kamelet

`script/validator/validator.go` gains one line — this Kamelet joins the `verifyUsedParams` exemption list, for exactly the reason `aws-s3-event-based-source` is already on it: `getObject` is consumed by a build-time `precondition`, which the declared-parameter check cannot see.

That list growing by hand is the smell I noted on #1173. I have followed the existing mechanism rather than changing the validator's contract in a Kamelet PR, but it is one more argument for replacing the hardcoded paths with an annotation the Kamelet itself carries.

## Open question for @yuvalif

The Ceph extension fields — `s3.bucket.id`, `s3.object.metadata`, `s3.object.tags`, and the record-level `eventId` and `opaqueData` — are passed through in the body but not surfaced as headers. `eventId` in particular is documented as usable for acking, so it would make a natural idempotency key. Happy to add them as headers in a follow-up if that is useful; I did not want to guess at which ones matter.

---


yuvalif:
@oscerd thanks for doing that!

> ## Open question for @yuvalif
> The Ceph extension fields — `s3.bucket.id`, `s3.object.metadata`, `s3.object.tags`, and the record-level `eventId` and `opaqueData` — are passed through in the body but not surfaced as headers. `eventId` in particular is documented as usable for acking, so it would make a natural idempotency key. Happy to add them as headers in a follow-up if that is useful; I did not want to guess at which ones matter.
> 
> _Claude Code on behalf of Andrea Cosentino_

most of these extensions would be useful 
* `s3.object.metadata` this is user information added as attributed on the object. in Ceph you can use these attributes for notification filtering. in addition, if the kamelet is used in order to consume these events into a database (or a catalog system), metadata could be useful
* `s3.object.tags`: similar to metadata but more structured
* `eventId`: would be used for finding duplicate events. this may happen due to retransmits in case of kafka disconnects
* `opaqueData`: extra user info associated with the topic. used for customizations that we could not predict beforhand. from my experience, used very creatively by users to work around issues :-)
* `s3.bucket.id` - probably not needed. this is an internal ceph ID of the bucket. useful for debugging (if you delete a bucket and create it with the same name, it will have a different ID)

oscerd:
Thanks @yuvalif — that is exactly the detail I needed. Pushed, following your calls one for one.

## Added as headers

| field | header | why, per your comment |
|---|---|---|
| `s3.object.metadata` | `ceph-object-metadata` | user attributes; usable for notification filtering, and useful feeding a catalog or database |
| `s3.object.tags` | `ceph-object-tags` | same but more structured |
| `eventId` | `ceph-event-id` | spotting duplicates after a Kafka retransmit |
| `opaqueData` | `ceph-opaque-data` | user information on the topic |

`s3.bucket.id` left out, as you suggested — internal id, debugging only.

Each also gets a `ce-` CloudEvents counterpart (`ce-cepheventid` and so on), which is how this catalog names source output headers — `infinispan-source` and `ftp-source` both do the plain-plus-`ce-` pair. That keeps the JSON paths out of the consumer's way.

Verified with a populated payload:

```
id=[evt-12345] opaque=[me@example.com]
meta=[[{key=x-amz-meta-owner, val=alice}]] tags=[[{key=env, val=prod}]]
```

`metadata` and `tags` arrive as structured lists, not flattened strings, so your "more structured" point survives into the consumer.

## One thing I would like your read on

You named `eventId` for duplicate detection, which is the use I had hoped for too. But the event structure in the Ceph docs carries it **empty**:

```json
"eventId":"",
```

and with that exact payload the header is set but blank:

```
EDGE id=[] opaque=[] meta=[[]] tags=[[]]
```

So anyone deduplicating on `ceph-event-id` would get a header that is always equal and silently dedupe nothing — the worst kind of failure, because it looks like it is working.

Is `eventId` populated in practice, and does it depend on something — persistent topics, a particular RGW version? If it is only populated in some configurations, I would rather say so in the Kamelet description than let people build idempotency on it and find out later.

## One behaviour change worth flagging

Reading the `Records` array means a message that is **not** a Ceph notification now fails the exchange:

```
PathNotFoundException: Missing property in path $['Records']
```

Previously, with `getObject` off, the Kamelet passed anything through untouched. I kept the strict behaviour deliberately — silently emitting blank Ceph headers for foreign messages would be worse than failing — and the description now says the topic is expected to carry only Ceph notifications. Happy to relax it if you think a shared topic is a realistic deployment.

---

PR #3030, opened 2026-09-14
Fix #2303: surface the Ceph extension fields as headers

### LINKED ISSUE #2303: Ceph: support bucket notifications?

oscerd:
Re-lands work that **missed the merge of #3026**. That PR was merged at its first commit, so the header change never reached `main` — `grep ceph-event-id kamelets/ceph-event-based-source.kamelet.yaml` returns nothing there today.

## What it adds

Following @yuvalif's field-by-field guidance on #3026:

| field | header | his reasoning |
|---|---|---|
| `s3.object.metadata` | `ceph-object-metadata` | user attributes; Ceph can filter notifications on them, and they matter feeding a catalog or database |
| `s3.object.tags` | `ceph-object-tags` | same but more structured |
| `eventId` | `ceph-event-id` | spotting duplicates after a Kafka retransmit |
| `opaqueData` | `ceph-opaque-data` | user information attached to the topic |

`s3.bucket.id` deliberately left out — he called it an internal id, useful for debugging only.

Each gets a `ce-` CloudEvents counterpart, which is how this catalog names source output headers; `infinispan-source` and `ftp-source` both do the plain-plus-`ce-` pair.

## Verified against a real notification

@yuvalif [posted a notification captured from his deployment](https://github.com/apache/camel-kamelets/pull/3030#issuecomment-5664886232). Running it through this branch:

```
id=[1789392761.006254.5eb63bbbe01eeed093cb22bb8f5acdc3]
opaque=[my-opaque-data-value-123]
meta=[[{key=x-amz-meta-author, val=yuval}, {key=x-amz-meta-project, val=ceph-rgw}]]
tags=[[{key=env, val=test}, {key=team, val=rgw}]]
ce-id=[1789392761.006254.5eb63bbbe01eeed093cb22bb8f5acdc3]
```

All four fields populate, and `metadata` and `tags` arrive as structured lists rather than flattened strings.

That capture also settled the question this PR originally left open. **`eventId` is populated in practice** — `<timestamp>.<counter>.<etag>`, unique per event — so it does work for deduplication. The empty `eventId` in the Ceph documentation is a placeholder in the sample, not what RGW sends, and the partial no longer warns about it. The same capture shows `metadata` and `tags` in an identical key/val shape, so calling tags "more structured than the metadata" was misleading; that is corrected too.

## Behaviour on a foreign message

Reading the `Records` array means a message that is not a Ceph notification fails the exchange:

```
PathNotFoundException: Missing property in path $['Records']
```

Previously, with `getObject` off, anything passed through untouched. The strict behaviour is deliberate — emitting blank Ceph headers for foreign messages would be worse than failing — and the partial states that the topic is expected to carry only Ceph notifications.

A missing extension field fails the same way, and that is correct rather than a gap: the [documented event structure](https://docs.ceph.com/en/latest/radosgw/notifications/#events) carries `eventId`, `opaqueData`, `metadata` and `tags` always, empty when nothing is set. I checked the inverse payload with those keys deleted, and it raises `PathNotFoundException` — but such a payload is not a Ceph event.

## Checks

`script/validator` reports no errors, the doc generator produces no diff beyond the partial, and `mvn clean install` passed with tests from the repository root on the previous revision. The latest commit is documentation only.

---


yuvalif:
captured JSON with eventId, metadata and tags.
```json
{
    "Records": [
        {
            "eventVersion": "2.2",
            "eventSource": "ceph:s3",
            "awsRegion": "default",
            "eventTime": "2026-09-14T13:32:41.001046Z",
            "eventName": "ObjectCreated:Put",
            "userIdentity": {
                "principalId": "testid"
            },
            "requestParameters": {
                "sourceIPAddress": ""
            },
            "responseElements": {
                "x-amz-request-id": "tx00000328fd24f363230a4-006aa7f776-4179-default",
                "x-amz-id-2": "4179-default-default"
            },
            "s3": {
                "s3SchemaVersion": "1.0",
                "configurationId": "notif-demo-config",
                "bucket": {
                    "name": "notif-demo-bucket",
                    "ownerIdentity": {
                        "principalId": "testid"
                    },
                    "arn": "arn:aws:s3:default::notif-demo-bucket",
                    "id": "c72557fd-8ecf-481f-b384-a50e0db94845.4179.1"
                },
                "object": {
                    "key": "hello.txt",
                    "size": 11,
                    "eTag": "5eb63bbbe01eeed093cb22bb8f5acdc3",
                    "versionId": "",
                    "sequencer": "79F7A76A92715F00",
                    "metadata": [
                        {
                            "key": "x-amz-meta-author",
                            "val": "yuval"
                        },
                        {
                            "key": "x-amz-meta-project",
                            "val": "ceph-rgw"
                        }
                    ],
                    "tags": [
                        {
                            "key": "env",
                            "val": "test"
                        },
                        {
                            "key": "team",
                           "val": "rgw"
                        }
                    ]
                }
            },
            "eventId": "1789392761.006254.5eb63bbbe01eeed093cb22bb8f5acdc3",
            "opaqueData": "my-opaque-data-value-123"
        }
    ]
}
```

oscerd:
Thanks @yuvalif — that capture is exactly what was needed, and it settled the one thing this PR had to leave open.

Running your payload through the branch:

```
id=[1789392761.006254.5eb63bbbe01eeed093cb22bb8f5acdc3]
opaque=[my-opaque-data-value-123]
meta=[[{key=x-amz-meta-author, val=yuval}, {key=x-amz-meta-project, val=ceph-rgw}]]
tags=[[{key=env, val=test}, {key=team, val=rgw}]]
ce-id=[1789392761.006254.5eb63bbbe01eeed093cb22bb8f5acdc3]
```

All four headers populate, and `metadata` and `tags` come through as structured lists rather than flattened strings.

**On `eventId`.** I had written a caveat into the documentation saying Ceph may emit it empty, so deduplicating on it might silently do nothing — taken from the event structure in your docs, which shows `"eventId":""`. Your capture shows it is populated, `<timestamp>.<counter>.<etag>` and unique per event, so the caveat was wrong and the empty value in the docs sample is just a placeholder. Removed it, and the documentation now describes the real format.

**One correction to my own wording.** I had described tags as being "in a more structured form than the metadata", following your note on #3026. Your capture shows both arrive in the same `[{key, val}]` shape, so that was misleading and is now fixed.

**One thing your capture does not cover, which I checked separately.** Your object had metadata, tags and opaque data all set. I built the inverse — those three keys deleted — and the route fails with `PathNotFoundException` rather than emitting empty headers. I did not treat that as a bug to fix, because the documented event structure carries all four keys always and merely leaves them empty when nothing is set, which your capture is consistent with. So a payload missing them is not a Ceph notification, and failing is the same behaviour the Kamelet already has for any foreign message on the topic.

If RGW can in fact omit those keys in some configuration or older release, that assumption is wrong and the JSONPath steps should suppress the exception instead — worth saying so if you know of such a case.

Documentation-only change on top of the reviewed commit; the Kamelet template is unchanged.

---

Issue #2868, opened 2026-06-22
camel-kamelets - itest failures running locally


davsclaus:
[ERROR] Errors:
[ERROR]   AwsIT.aws()[10] » TestCaseFailed Action timeout after 120000 milliseconds.
[ERROR]   AwsIT.aws()[11] » TestCaseFailed Action timeout after 120000 milliseconds.
[ERROR]   AwsIT.aws()[12] » TestCaseFailed AWS DDB item verification failed
[ERROR]   AwsIT.aws()[8] » TestCaseFailed AWS DDB item verification failed
[ERROR]   CommonIT.crypto()[1] » TestCaseFailed Action timeout after 60000 milliseconds. Failed to receive message on endpoint: 'httpServer.inbound'
[ERROR]   CommonIT.filter()[1] » TestCaseFailed Action timeout after 60000 milliseconds. Failed to receive message on endpoint: 'httpServer.inbound'
[ERROR]   CommonIT.header()[1] » TestCaseFailed Action timeout after 60000 milliseconds. Failed to receive message on endpoint: 'httpServer.inbound'
[ERROR]   CommonIT.httpTests()[1] » TestCaseFailed Action timeout after 60000 milliseconds. Failed to receive message on endpoint: 'httpServer.inbound'
[ERROR]   CommonIT.httpTests()[2] » TestCaseFailed Action timeout after 60000 milliseconds. Failed to receive message on endpoint: 'httpServer.inbound'
[ERROR]   CommonIT.transformation()[1] » TestCaseFailed Action timeout after 60000 milliseconds. Failed to receive message on endpoint: 'httpServer.inbound'
[ERROR]   CommonIT.transformation()[2] » TestCaseFailed Action timeout after 60000 milliseconds. Failed to receive message on endpoint: 'httpServer.inbound'
[ERROR]   CommonIT.transformation()[3] » TestCaseFailed Action timeout after 60000 milliseconds. Failed to receive message on endpoint: 'httpServer.inbound'
[ERROR]   CommonIT.transformation()[4] » TestCaseFailed Action timeout after 60000 milliseconds. Failed to receive message on endpoint: 'httpServer.inbound'
[ERROR]   CommonIT.transformation()[5] » TestCaseFailed Action timeout after 60000 milliseconds. Failed to receive message on endpoint: 'httpServer.inbound'
[...]

PR #3032, opened 2026-09-15
Fix #2871: surface Kafka record metadata as non-internal headers

### LINKED ISSUE #2871: Kafka kamelets should use kamelet properties instead of leaking Camel internal headers
## Summary

The Kafka kamelets currently rely on Camel-internal headers (`CamelKafkaTopic`, `CamelKafkaKey`, `CamelKafkaPartition`, `CamelKafkaTimestamp`) being forwarded across transport boundaries (e.g. HTTP → Kafka). This is a design issue:

1. **Header leaking**: Previously, old-style `kafka.TOPIC` / `kafka.KEY` headers could leak through HTTP because they weren't filtered by the HTTP component's `DefaultHeaderFilterStrategy`. With the migration to `CamelKafka*` names, the HTTP component now correctly filters them — but this broke tests that relied on the leak.

2. **Wrong abstraction level**: If a kamelet needs to let users dynamically control the Kafka topic, key, or partition, that should be exposed as a kamelet-level property (e.g. `kafkaTopic`) declared in `spec.definition.properties`, not by requiring the caller to set a Camel-internal header directly. The kamelet template should map the property to the internal header.

## Current behavior

- `kafka-sink` expects the caller to set `CamelKafkaKey` / `CamelKafkaPartitionKey` headers directly (or via `key` / `ce-key` CloudEvents headers that the template maps)
- `kafka-source` exposes `CamelKafkaTopic` / `CamelKafkaKey` / `CamelKafkaPartition` as outgoing headers that leak to downstream consumers
- `timestamp-router-action` and `message-timestamp-router-action` read `CamelKafkaTopic` / `CamelKafkaTimestamp` from the exchange — the caller must set these as Camel-internal headers
- Tests relied on `kafka.TOPIC` headers passing through HTTP, which was a leak, not a feature

## Proposed fix

- Add kamelet-level properties (e.g. `kafkaTopic`, `kafkaKey`, `kafkaPartition`) to the relevant kamelets
- The kamelet template maps these properties to the internal `CamelKafka*` headers
- For source kamelets, map `CamelKafka*` headers to non-internal output headers (e.g. `kafka-topic`, `kafka-key`) so downstream consumers can access them without depending on Camel internals
- Update tests to use the kamelet properties / non-internal headers
- Update the `timestamp-router-action` to accept a kamelet property for the topic instead of reading a Camel-internal header directly

This aligns with the kamelet design principle that the template is the trust boundary — callers should interact through declared properties, not by manipulating Camel-internal headers.

## Related

- #2869 — optional parameter endpoint creation failure
- #2868 — integration test failures (header migration)

oscerd:
Closes the one part of #2871 that was still outstanding.

## Why this is the only part left

Most of #2871 was overtaken by #2868, which landed after it was filed:

| #2871 proposed | state on `main` |
|---|---|
| `timestamp-router-action` should take a property for the topic rather than read a Camel internal header | done — `topicHeaderName` / `timestampHeaderName` (622edd686) |
| Callers should not have to set `CamelKafka*` on sinks | done — `kafka-sink` maps `key` / `ce-key` → `CamelKafkaKey`, `partition-key` / `ce-partitionkey` → `CamelKafkaPartitionKey` |
| The `kafka.TOPIC` through HTTP leak in tests | done (2e373d5d6, 49d30bc75) |
| **Sources should map `CamelKafka*` to non-internal output headers** | **nothing done — this PR** |

I did not implement the issue's other suggestion, exposing `kafkaTopic` / `kafkaKey` / `kafkaPartition` as Kamelet **properties**. A Kamelet property binds once when the template is instantiated, so it cannot carry a per-message topic, key or partition, which is exactly what those values are. The friendly-header indirection `kafka-sink` already uses is the abstraction that works for per-message values, and it is already in place. Reasoning posted on the issue.

## What this adds

```
kafka-topic      ce-kafkatopic      from CamelKafkaTopic
kafka-key        ce-kafkakey        from CamelKafkaKey
kafka-partition  ce-kafkapartition  from CamelKafkaPartition
kafka-offset     ce-kafkaoffset     from CamelKafkaOffset
kafka-timestamp  ce-kafkatimestamp  from CamelKafkaTimestamp
```

Plain name plus a CloudEvents one, which is how this catalog names source output headers — `infinispan-source` and `ftp-source` are the pattern followed here.

The names were taken from the `camel-kafka` catalog in 4.23.0-SNAPSHOT rather than assumed. Only `consumer` group headers are surfaced; `CamelKafkaHeaders`, `CamelKafkaManualCommit`, `CamelKafkaLastPollRecord` and `CamelKafkaLastRecordBeforeCommit` are flow control rather than record metadata, so they are left alone.

**Purely additive.** The `CamelKafka*` headers stay on the exchange, so anything already reading them keeps working.

## Verified

`KafkaIT` run locally against a real Redpanda broker, with the new assertions in place:

```
✔ SUCCESS (52038ms) kafka-sink-pipe-test
✔ SUCCESS (31308ms) kafka-source-pipe-test
✔ SUCCESS (24270ms) kafka-router-pipe-test
TOTAL: 3   PASSED: 3 (100.0%)
```

`kafka-source-pipe-test` now asserts the new headers arrive at the HTTP sink — exact values for `kafka-topic` and `ce-kafkatopic`, `@notEmpty()@` for key, partition, offset and timestamp, since those are not fixed.

Also probed the missing-header case directly: a record with no key sets `kafka-key` empty rather than failing the exchange.

`script/validator` reports no errors, the doc generator produces no diff beyond the partial, and `KameletsCatalogTest` passes (18 tests).

## Note for a separate PR

While editing `kafka-source-description.adoc` I found its Authentication and Configuration sections are wrong, independently of this change: they document **User** and **Password** as required properties and claim the Kamelet "requires SASL/PLAIN authentication". The actual properties are `saslUsername` / `saslPassword`, both optional, with `saslAuthType` defaulting to `NONE` — and both usage examples in that file pass `user:` / `password:`, which are not properties at all. I corrected only the Output Format section here, since that is what this change affects. The rest deserves its own fix.

---


davsclaus (approved):
LGTM, thanks. The mapping follows the `infinispan-source` pattern exactly, it is purely additive so nothing reading `CamelKafka*` breaks, and the strengthened `kafka-source-pipe-test` (exact match on topic, `@notEmpty()@` on the rest) pins it down. The scope table explaining why this is the only part of #2871 still outstanding is much appreciated.

A few non-blocking notes:

1. **Sibling Kafka sources.** Only `kafka-source` gets the mapping. The apicurio / azure-schema-registry / not-secured source variants consume the same per-record `CamelKafka*` headers and would benefit from the same block (the batch sources are a different story since the body is a list). Fine as a follow-up, but worth a note on #2871 so it is not lost.
2. **Null key.** For a record without a key, `setHeader kafka-key` stores a null-valued header rather than omitting it. Harmless, and `infinispan-source` behaves the same way, so consistency argues for leaving it as is; just flagging in case an absent header is preferred.
3. **Source/sink naming asymmetry.** The source now emits `kafka-key`, while `kafka-sink` reads `key` / `ce-key`. In a source-to-sink pipe the `CamelKafkaKey` pass-through still carries the key, so nothing breaks, but the friendly names do not round-trip. Might be worth a `kafka-key` mapping on the sink later.

_This review was generated by an AI agent and may contain inaccuracies. Please verify all suggestions before applying._

Issue #1888, opened 2024-02-13
Rework the scripts as maven plugin


claudio4j:
Rework the generator.go and validator.go from the scripts directory as kamelet maven plugin to facilitate maintenance and consistency in the project regarding the primary programming language as java.

claudio4j:
The intention is to move the go code to the existing maven plugin and create a new generator mojo. Because today, we have to run the maven plugin and the go scripts when there are changes in the kamelets.

PR #3046, opened 2026-09-17
Fix #1888: replace the Go doc generator with a Maven mojo

oscerd:
First half of #1888 — @claudio4j's point that changing a Kamelet currently means running Maven *and* a separate Go toolchain.

## The generator is much smaller than it looks

`generator.go` is 105 lines but writes exactly **one** file: `docs/modules/ROOT/nav.adoc`, an `xref` per Kamelet between two generated-file markers. The per-Kamelet pages it once emitted were removed in 6b7d9d520, so most of the file is scaffolding around a small job.

`GenerateNavMojo` does the same: read the kamelets directory, sort by file path, take `metadata.name` from each, write the file.

## Wiring

Bound to `compile` in `camel-kamelets-bom`, alongside the existing `validate` goal, so an ordinary build regenerates the navigation. Both executions now carry ids, since there is more than one.

Jackson does the YAML parsing and is **not** declared in the plugin pom — `ValidateKameletsMojo` already uses Jackson the same way. I did try adding it explicitly, but the version is not managed anywhere in this build, and pinning one here would risk drifting from whatever Camel pulls in.

## Verified byte for byte

The check that matters for a rewrite like this:

1. generate `nav.adoc` with the Go tool, keep it
2. delete the file
3. build
4. `cmp`

```
IDENTICAL (9078 bytes, 264 lines)
```

Then a full `mvn clean install` from the root: `nav.adoc` comes out identical again and does not show up in `git status`, so the build is idempotent. `KameletsCatalogTest` passes (18 tests) and the validator still reports no errors.

## CI

With the generator gone, `main-push-regen.yaml` has no remaining use for Go, so both the `setup-go` step and the `Run Generator` step are removed. The two `./mvnw clean install` runs it already performs now produce `nav.adoc` themselves, so the regen bot keeps working the same way — it just picks the file up from the build instead of a separate step.

I checked `dependabot.yml` for a gomod entry pointing at the deleted directory; there is none, it only watches `/` for maven and github-actions.

`README.md` and `AGENTS.md` are updated to match.

## Not in this PR

`validator.go` is 506 lines and is the larger half of #1888. It stays in Go here and is untouched, including its `validate.yaml` workflow. Worth landing separately once this one is trusted.

---


davsclaus (approved):

claudio4j (changes requested):
There is a change to do in Makefile "generate" target, it calls the deleted go code. Also there a couple of mentions of script/generator in the ./docs/modules/ROOT/pages/security-model.adoc and AGENTS.md.

oscerd:
Thanks @claudio4j — all three fixed in 90a1049ad.

**Makefile `generate`.** It invoked the deleted code, and removing those two lines also had to fix the step after them: the target used to `cd ../script/generator`, then reach back with `cd ../../crds`. With the generator gone it stays in `crds` and calls `./gen_crd/gen_crd_api.sh` directly. Checked with `make -n generate`. I added a note that `nav.adoc` comes from the Maven build now.

```make
# nav.adoc is generated by the Maven build, see the install target
generate:
	cd crds && \
	./gen_client.sh && \
	./generate.sh && \
	./gen_crd/gen_crd_api.sh
```

**The two stale descriptions.** `AGENTS.md` said Go is required for the "`script/` and `crds/` generators" and described `script/` as the "Go doc generator + YAML validator"; `security-model.adoc` listed `script/` as "(Go generator/validator)". All now say validator.

**Why I missed them.** My sweep searched for the literal string `script/generator` restricted to Markdown and YAML. That matches none of the three: the Makefile was not in the file types I searched, and both prose mentions describe the directory without naming the path. I have re-run it as a case-insensitive search for the path, `generator.go`, "doc generator" and "generator/validator" across every file type. What is left is the CHANGELOG, which is release history and should not be rewritten, and the new mojo's own javadoc line saying which Go code it replaces.

Nothing about the generated output changed: `nav.adoc` is still 264 lines and 9078 bytes and does not appear in `git status` after a full build.

---

claudio4j (approved):

